PR #24, opened 2021-06-12
Add document toolbar icons

Paula-Kli:
In https://github.com/fsr-de/1327 there is also an icon for adding emojis - Do you think we want this again? I have the feeling it was not used...

### DIFF
--- a/myhpi/settings.py
+++ b/myhpi/settings.py
@@ -156,7 +156,7 @@ STATICFILES_STORAGE = "django.contrib.staticfiles.storage.ManifestStaticFilesSto
 
 STATIC_URL = env.str("STATIC_URL")
 STATIC_ROOT = env.str("STATIC_ROOT")
-STATICFILES_DIRS = (os.path.join(BASE_DIR, "static"),)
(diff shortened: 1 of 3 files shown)


myhpi/settings.py
```diff
@@ -156,7 +156,7 @@
 
 STATIC_URL = env.str("STATIC_URL")
 STATIC_ROOT = env.str("STATIC_ROOT")
-STATICFILES_DIRS = (os.path.join(BASE_DIR, "static"),)
+STATICFILES_DIRS = (os.path.join(BASE_DIR, "myhpi/static"),)
```
Paula-Kli, line 159: @jeriox Do you have a clue why I needed to add this in order to find the static files with `python manage.py collectstatic`?

T4rikA (approved):

myhpi/settings.py
jeriox: oh yeah it was just the wrong path in the settings file. my fault, didn't change that when we refactored stuff a while ago. It's not necessary to collect the static files for local development, therefore this did not stand out. thx for the catch

jeriox (approved):
looks good, thx!

PR #29, opened 2021-12-16
Add OIDC login

### LINKED ISSUE #4: OIDC Login

T4rikA:
closes #4 

- [x]  Reminder: Add changes in env file to README as soon as #27 is merged

### DIFF
--- a/myhpi/templates/login.html
+++ b/myhpi/templates/login.html
@@ -1,0 +1,9 @@
+{% if user.is_authenticated %}
+  <p>Current user: {{ user.email }}</p>
+  <form action="{% url 'oidc_logout' %}" method="post">
+    {% csrf_token %}
+    <input type="submit" value="logout">
+  </form>
+{% else %}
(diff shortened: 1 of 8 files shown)


myhpi/templates/login.html
```diff
@@ -0,0 +1,9 @@
+{% if user.is_authenticated %}
```
Paula-Kli, line 1: what does `is_authenticated` mean? for me it reads like: "I can login if I am already authenticated" ?

Paula-Kli:
As much as I understand it it looks fine to me :+1:

myhpi/templates/login.html
T4rikA: Exactly, but this is a really good catch because the template is not used yet 😅

Paula-Kli: what is it for then?

T4rikA: It should be used, we just forgot to finish this

jeriox (approved):

PR #30, opened 2022-05-01
Update dependencies

Paula-Kli:
start of updating dependencies but wagtail itself is still missing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,6 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
-    rev: 21.4b2
+    rev: 22.3.0
     hooks:
       - id: black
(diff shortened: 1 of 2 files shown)


T4rikA:
I also updated the dependencies in #29, is there a reason why you didn't update wagtail?

Paula-Kli:
> I also updated the dependencies in #29, is there a reason why you didn't update wagtail?

because I did not know that you would update the dependencies in #29 since it is about OIDC and I did not update wagtale since I did not have time yet und would update it as soon as possible

jeriox:
I removed the updates from #29, but you still need to rerun `poetry lock` because we installed an additional dependency

frcroth:
@Paula-Kli Can this be closed?

Paula-Kli:
> @Paula-Kli Can this be closed?

I think so

Issue #10, opened 2021-04-24
As a user I want to see my status groups if I hover over my name


T4rikA:
+ logout button

PR #31, opened 2022-05-01
Adds a rudimentary user interface

ksturtzkopf:


ksturtzkopf:
Fixes #10

lukasrad02:
`href` for Login/Logout is still missing, we have to figure out the correct endpoints.

And the strings are not localized yet.

jeriox:
@lukasrad02 the correct endpoints are now available as of #29

jeriox:
@ksturtzkopf could you please merge/rebase main and resolve the merge conflicts? will review afterwards

frcroth:
@ksturtzkopf Are you still working on this?

myhpi/static/js/myHPI.js
```diff
@@ -21,3 +21,14 @@ function addCollapseToLevel2NavContainers() {
...
+
+// POST Request to handle logging out
+window.addEventListener("load",function() {
+    if (document.getElementById("navbarLogoutLink") !== null) {
+      document.getElementById("navbarLogoutLink").addEventListener("click",function(e) {
+        e.preventDefault();
+        document.getElementById("logoutForm").submit();
+      });
+    }
+  });
```
frcroth, line 34: \n

myhpi/static/js/myHPI.js
```diff
@@ -21,3 +21,14 @@ function addCollapseToLevel2NavContainers() {
...
 }
+
+
+// POST Request to handle logging out
+window.addEventListener("load",function() {
+    if (document.getElementById("navbarLogoutLink") !== null) {
+      document.getElementById("navbarLogoutLink").addEventListener("click",function(e) {
+        e.preventDefault();
+        document.getElementById("logoutForm").submit();
+      });
```
frcroth, line 32: Why do we need these verbose event listeners? Wouldn't a reference in the onclick attribute of the link suffice?

myhpi/static/scss/myHPI.scss
```diff
@@ -140,3 +140,19 @@ $colors: map-merge($colors, $custom-colors);
...
+
+#userGroupsHeader {
+    font-style: italic;
+    &:hover {
+        color: $dropdown-link-color;
+        background-color: transparent; // For `<button>`s
+    }
+}
+
+#userMenu {
```
frcroth, line 156: Follow kepab-case

myhpi/templates/base.html
```diff
@@ -34,25 +34,53 @@
...
-                    <img src="{% static 'img/FSR_DE_color.svg' %}" class="w-100" alt="Zur Startseite">
-                </a>
-                {% build_menu_level sub_pages_by_level_and_id 0 %}
+    <nav class="navbar-myhpi container-lg d-print-none page">
+        <div class="bg-white">
+            <div class="navbar-top d-flex flex-between page-content">
+                {% with all_pages|all_sub_menu_pages_by_id as sub_pages_by_level_and_id %}
+                <div class="d-flex flex-between">
+                    <a class="nav-brand" href="{% pageurl root_page %}">
+                        <img src="{% static 'img/FSR_DE_color.svg' %}" class="w-100" alt="Zur Startseite">
```
frcroth, line 43:
```suggestion
                        <img src="{% static 'img/FSR_DE_color.svg' %}" class="w-100" alt="To home page">
```

myhpi/templates/base.html
```diff
@@ -34,25 +34,53 @@
...
+                <div class="align-self-center ms-auto">
+                    {% if user.is_anonymous %}
+                        <a href="{% url 'oidc_authentication_init' %}{% if 'next' in request.GET %}?next={{ request.GET.next }}{% endif %}">Login</a>
+                    {% else %}
+                    <div class="dropdown">
+                        <a href="#" class="btn dropdown-toggle" data-bs-toggle="dropdown" aria-expanded="false" id="userMenu">
+                            {{ user.get_full_name }}
+                        </a>
+                        <ul class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+                            <li class="dropdown-item" id="userGroupsHeader">Deine Gruppen</li>
```
frcroth, line 56:
```suggestion
                            <li class="dropdown-item" id="userGroupsHeader">Your groups</li>
```

frcroth (changes requested):
About the commit messages: Drop the s from the verb and also "Rebase" is no useful commit message. If you rebase, you usually rebase commits, but there is never a rebase-commit - that's the beauty of it.

myhpi/templates/base.html
jeriox: we could already use the translate template tags here. needs to be adapted soon anyways

ksturtzkopf: Will wagtail provide the details of the translated strings automatically (via settings or similar), or do we need to implement that ourselves?

jeriox: the translate tag is a built-in function of django, see https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#translate-template-tag.
The translations are stored in a specific file in the repository and django takes care of displaying the correct string for the current language

ksturtzkopf: Is there a particular style you wish for me to follow or is any style I deem fit fine?

myhpi/core/models.py
```diff
@@ -56,8 +50,16 @@ class InformationPage(BasePage):
...
+
+    @property
+    def last_edited_by(self):
+        return self.get_latest_revision().user
+
+    settings_panels = [
+        PublishingPanel(),
+        FieldPanel("is_public", widget=forms.CheckboxInput),
+        FieldPanel("visible_for", widget=forms.CheckboxSelectMultiple),
+        FieldPanel("author_visible", widget=forms.CheckboxInput(attrs={"checked": ""})),
```
frcroth, line 62: Remove this and bring the search fields back

myhpi/core/models.py
```diff
@@ -30,11 +29,6 @@ class BasePage(Page):
         FieldPanel("is_public", widget=forms.CheckboxInput),
         FieldPanel("visible_for", widget=forms.CheckboxSelectMultiple),
     ]
-    # FilterFields required for restricting search results
-    search_fields = Page.search_fields + [
-        index.FilterField("group_id"),
-        index.FilterField("is_public"),
-    ]
```
frcroth, line 32: Search fields lost on merge

myhpi/core/models.py
```diff
@@ -168,11 +170,6 @@ class Minutes(BasePage):
     ]
     parent_page_types = ["MinutesList"]
     subpage_types = []
-    search_fields = BasePage.search_fields + [
-        index.SearchField("text"),
-        index.SearchField("participants"),
-        index.SearchField("moderator"),
-    ]
```
frcroth, line 173: Search fields  lost on merge

myhpi/polls/models.py
```diff
@@ -51,10 +50,6 @@ class Poll(BasePage):
         "PollList",
     ]
     subpage_types = []
-    search_fields = BasePage.search_fields + [
-        index.SearchField("description"),
-        index.SearchField("question"),
-    ]
```
frcroth, line 57: Search fields lost on merge

myhpi/search/templates/search/search.html
```diff
@@ -28,5 +35,4 @@ <h4><a href="{% pageurl result %}">{{ result }}</a></h4>
     {% elif search_query %}
         No results found
     {% endif %}
```
frcroth, line 37: This file should not have been changed

myhpi/search/templates/search/search_field.html
```diff
@@ -1,10 +0,0 @@
-
-
-<form action="{% url 'search' %}" method="get">
-    <div class="input-group">
-        <div class="form-outline">
-            <input type="search" class="form-control" name="query"{% if search_query %} value="{{ search_query }}"{% endif %}>
-        </div>
-        <button type="submit" class="btn btn-primary">Search</button>
-    </div>
-</form>
```
frcroth, line 10: This file should not have been removed

myhpi/search/views.py
```diff
@@ -1,26 +1,16 @@
...
     if search_query:
-        user_groups = get_user_groups(request.user)
-        allowed_pages = (
-            BasePage.objects.live()
-            .filter(Q(visible_for__in=user_groups) | Q(is_public=True))
-            .distinct()
-        )
-        search_results = allowed_pages.search(search_query)
+        search_results = Page.objects.live().search(search_query)
         query = Query.get(search_query)
```
frcroth, line 15: This file should not have been changed

tests/test_minutes_extensions.py
```diff
@@ -1,6 +1,9 @@
+import os
+
 import django
 from django.test import TestCase
 
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "myhpi.settings")
```
frcroth, line 6: This line is not required for me

frcroth (changes requested):
Your merge removed all stuff related to search :(

frcroth (commented):
The username and status groups are not currently shown, right? Do you want to include this as specified in the original issue?
Logout/login seems fine

myhpi/static/js/myHPI.js
```diff
@@ -63,3 +63,9 @@ window.onresize = () => {
     adaptNavbarToWindowSize()
     updateNavbarPosition()
 }
+
+// POST Request for logging out
+document.getElementById("navbar-logout-link").addEventListener("click",function(e) {
```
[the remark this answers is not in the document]
ksturtzkopf: To me this is more canonical, but I'd be more than happy to change it.

ksturtzkopf:
> The username and status groups are not currently shown, right? Do you want to include this as specified in the original issue? Logout/login seems fine

Do you have a design specification or should I just implement that as I see fit?

frcroth:
> > The username and status groups are not currently shown, right? Do you want to include this as specified in the original issue? Logout/login seems fine
> 
> Do you have a design specification or should I just implement that as I see fit? @frcroth

You can show the status groups on hover and maybe the user name always, but if you have better ideas go ahead

myhpi/static/js/myHPI.js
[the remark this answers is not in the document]
jeriox: I just personally don't like to do stuff in JS if the plain HTML variant works just as good ^^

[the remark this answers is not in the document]
ksturtzkopf: Now implemented as a plain HTML-form.

myhpi/templates/base.html
```diff
@@ -76,13 +81,20 @@
                 <div id="nav-level-2" class="nav-level"></div>
 
                 <div id="nav-level-right"
-                    class="nav-level d-flex flex-column flex-grow-1 align-items-stretch align-items-xl-end">
+                    class="nav-level d-flex flex-column flex-grow-2 align-items-stretch align-items-xl-end">
                     <ul id="nav-item-container-user" class="nav-item-container collapse"
                         data-navbar-level="#nav-level-0">
+                        {% if user.groups.all|length > 0 %}<span class="nav-bar-group-header"><i>Deine Gruppen:</i></span>{% endif %}
```
jeriox, line 87: `if user.groups.exists`

myhpi/templates/base.html
```diff
@@ -76,13 +81,20 @@
...
+                        {% for group in user.groups.all %}
+                            <li class="nav-bar-group-item">{{ group }} </li>
+                            {% empty %}
```
jeriox, line 90: what do you need the empty for?

ksturtzkopf: Removed in update. 
If we want to display something along the lines of "You're not part of any group yet, join now" etc. in the future we could use empty.

myhpi/templates/base.html
ksturtzkopf: Changed in update.

jeriox (approved):

jeriox:
thank you very much, lgtm now!

myhpi/static/scss/myHPI.scss
```diff
@@ -262,4 +262,11 @@ h1.toc-title::after {
         margin-top: 50px;
         order: 3;
     }
+}
+
+.nav-bar-group-header {
+    padding: $dropdown-item-padding-y $dropdown-item-padding-x;
+}
+.nav-bar-group-item {
+    padding: $dropdown-item-padding-y $dropdown-item-padding-x*2;
```
SilvanVerhoeven, line 271: Nav-related stuff can go into `navbar.scscc`

myhpi/templates/base.html
```diff
@@ -52,18 +52,23 @@
...
+                            {% bs_icon 'person-circle' size='1.4em' %}
+                        </a>
+                        <a id="nav-mobile-toggle" class="nav-link d-xl-none" href="#nav-item-container-root"
+                            onclick="toggleMobileNavbar(event)"
+                            role="button"
+                            aria-expanded="false" aria-controls="nav-item-container-root">
+                            {% bs_icon 'list' size='1.4em' %}
+                        </a>
+                    {% else %}
+                        <a id="nav-user-toggle" class="nav-link" class="dropdown-item" href="{% url 'oidc_authentication_init' %}{% if 'next' in request.GET %}?next={{ request.GET.next }}{% endif %}">Login</a>
```
SilvanVerhoeven, line 70: I'd prefer using "Anmelden" and "Abmelden" in German instead of "Login"/"Logout", but no hard opinion there

We could also use an icon plus tooltip "Login" instead of the text "Login" (imo having text next to the search icon looks a bit off). This boostrap icon might fit: [box-arrow-in-right](https://icons.getbootstrap.com/icons/box-arrow-in-right/).  
If we want to stick to the Login as text, we should make sure that there is no padding on the right, so the "Login" text is aligned with the website content below

myhpi/templates/base.html
```diff
@@ -76,13 +81,20 @@
                 <div id="nav-level-2" class="nav-level"></div>
 
                 <div id="nav-level-right"
-                    class="nav-level d-flex flex-column flex-grow-1 align-items-stretch align-items-xl-end">
+                    class="nav-level d-flex flex-column flex-grow-2 align-items-stretch align-items-xl-end">
```
SilvanVerhoeven, line 84: Why is this needed?

myhpi/templates/base.html
```diff
@@ -76,13 +81,20 @@
...
+                                {% endfor %}
+                            <li class="dropdown-divider"></li>
+                        {% endif %}
                         <li class="nav-item">
-                            <a class="nav-link" role="button">
-                                Abmelden
-                            </a>
+                            <form action="{% url 'oidc_logout' %}" method="post" id="logout-form">
+                                {% csrf_token %}
+                                <button type="submit" class="dropdown-item" href="/">Logout</button>
```
SilvanVerhoeven, line 96: Due to this being a button, the styling is completely off. Imo, the easiest solution is to use the link that was there before instead of the button tag and submit the form via JS like this: https://stackoverflow.com/questions/4286466/use-a-normal-link-to-submit-a-form

SilvanVerhoeven (changes requested):

myhpi/templates/base.html
jeriox: would just using the bootstrap `<button type="button" class="btn btn-link">` work? I'd really prefer not doing this in JS

ksturtzkopf: Would you mind sharing in which browser the styling seems off? I've tried to replicate the behaviour, however in Chromium based browsers, Safari and Firefox the styling is fine on my end.

myhpi/templates/base.html
ksturtzkopf: This might be a relict of changes no longer needed. Will be changed in the next commit.

myhpi/templates/base.html
ksturtzkopf: Agreed. This will be implemented with the localisation (#224). 

I will add the logo to improve the design.

myhpi/static/scss/myHPI.scss
ksturtzkopf: As of right now, only myHPI.scss appears to be imported into the base template. 
```$dropdown-item-padding``` is used in the bootstrap templates for dropdown items. 
Thus far i have found no occurrence of the class being overridden. Hence, ```$dropdown-item-padding```  is used (and results equal spacing across the different dropdown-items). 

Nevertheless, I will investigate to ensure proper behaviour :).

File: myhpi/templates/base.html (hunk 20 above)
Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: @ksturtzkopf In basically every browser (I am using Firefox). Could you run `python manage.py collectstatic` and reload the page? And then send in a screenshot of how it looks like?

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: On my end it looks like this (after running `python manage.py collecstatic`) 

Screenshot of myHPI landing page

Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: Might be the same as for me. Press the Logout button and have a look, whether it lights up in orange
grafik

Compare this to the look (e.g. colors and animations) from the other menu items
grafik

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: yes, the button does light up in orange when pressed. 

_However_, this appears to be the default behaviour for `dropdown-items`. On my end at least.

Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: May be. But our menu does not use `dropdown-items` from Bootstrap

Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: Compare with the other nav-items

Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: We *could* start using `dropdown-items` in the whole menu (and then restyle the Bootstrap `dropdown-item` classes). But on the other hand the menu is no dropdown in the bootstrap sense, so I am not sure whether we should really do this

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: I reckon it would make more sense to implement the desired behaviour for these items explicitly.
I agree that using JS is the more favourable solution (it was also the original implementation ;) ), but I fully understand why @jeriox prefers using plain HTML.
So, where do we go from here? 
I'd be happy to revert back to using JS and implement the styling using the classes available.

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: (and just as an fyi: Not using `dropdown-items` does result in the button being displayed as a button and not in the style we want to achieve ;) )

Comment (SilvanVerhoeven) [reply myhpi/templates/base.html:96]: >I reckon it would make more sense to implement the desired behaviour for these items explicitly.
Sorry, I what do you mean with "implementing explicitly"?

>(and just as an fyi: Not using `dropdown-items` does result in the button being displayed as a button and not in the style we want to achieve ;) )
As far as I can see neither adding nor omitting `dropdown-items` results in the style of the other menu items (but yes, adding `dropdown-items` makes it harder to spot)

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: As I would like to wrap this PR up: @jeriox @SilvanVerhoeven, I am more than happy to implement the Logout Routing in JS or leave it as a form. 
To prevent having large discussions after committing to one solution, I would like to ask you to mediate between you two as it seems that you have the strongest opinions on the matter.

Comment (ksturtzkopf) [reply myhpi/templates/base.html:96]: > Sorry, I what do you mean with "implementing explicitly"?

Perhaps that wasn't the best word to describe what I meant. 
To ensure that the design is matched I would restyle the dropdown menu to match the desired style and indentation without using the bootstrap components. (i.e. in `nav-link`)

Comment (jeriox) [reply myhpi/templates/base.html:96]: @SilvanVerhoeven convinced me that doing it without JS is too ugly and cumbersome regarding the design, so feel free to go ahead and implement it in JS :)

File: myhpi/static/js/navbar.js [hunk 21]
```diff
@@ -231,8 +231,8 @@ const moveRootLevelToDesktopLayout = () => {
     const rootNavLevelContainer = document.querySelector("#nav-level-0")
     const rootLevel = document.querySelector("#nav-item-container-root")
     rootNavLevelContainer.appendChild(rootLevel)
+    document.getElementById("user-information-username").hidden = false
```
SilvanVerhoeven, line 234: This breaks the website if no user is logged in, as no span with this ID exists in this situation.  
This could be fixed in JS, but I noticed that the use of the `hidden` attribute is [against the specs](https://stackoverflow.com/a/6708403/8325653).

So I'd opt for a CSS-only solution: Simply add the respective [Boostrap built-in class](https://getbootstrap.com/docs/5.0/utilities/display/) (with the appropriate breakpoint) to the span

SilvanVerhoeven (changes requested):

SilvanVerhoeven:
Looks good to me visually and works :)

SilvanVerhoeven (approved):

frcroth (approved):

PR #46, opened 2022-05-08
Create codeql-analysis.yml

jeriox:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,61 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '44 16 * * 0'
```
frcroth, line 10: Does this have any relevance? “At 16:44 on Sunday.”

jeriox: this was automatically created by GitHub after I pushed the corresponding button, I think they try to balance their load like this ^^

frcroth (approved):

PR #55, opened 2022-05-09
Update dependencies

Paula-Kli:
it is the same issue as https://github.com/fsr-de/myHPI/pull/30

The main changes are the updates of wagtail and Django. But also of Bleach.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -121,6 +121,7 @@ python-versions = ">=3.7"
 
 [package.dependencies]
 six = ">=1.9.0"
+tinycss2 = {version = ">=1.1.0", optional = true, markers = "extra == \"css\""}
 webencodings = "*"
 
 [package.extras]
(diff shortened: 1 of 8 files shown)


jeriox:
@Paula-Kli you can manually update wagtail-localize in pyproject.toml since we are several versions behind

jeriox:
If you don't want to search for the newest version, you can also run `poetry add wagtail-localize@latest` which will update the entry in pyproject.toml

poetry.lock
```diff
@@ -409,7 +424,7 @@ python-versions = "*"
 name = "isort"
 version = "5.10.1"
 description = "A Python utility / library to sort Python imports."
-category = "dev"
+category = "main"
```
Paula-Kli, line 427: for example here I changed the dependency from dev to main by accident but don't know how to change it back except for deleting isort and installing it new... does anybody have a good idea on how to change it back? I can not do it manually as poetry computes a hash over that file

jeriox (approved):

jeriox:
works for me as well, thx

PR #48, opened 2022-05-09
Feat/basic frontend

SilvanVerhoeven:
Close #9 

This is a basic Desktop only frontend for myHPI. Previous commits of this branch contain different proposals for the frontend. Current proposal:

grafik

This PR does not include following majaor features yet:

- Truly responsive UI
- Footer
- (Printing of) Minutes

There will be extra issues for these features

PR #57, opened 2022-05-16
Limited visibility pages and of menu items

### LINKED ISSUE #3: Restrict menu items to pages that the user is allowed to view

DrEGZo:
Resolves #3.

Pages can now be set to public, other than that their visibility can be limited to certain user groups.
Users without the necessary permissions to view a page will receive a 403.

Prior to rendering the navbar, pages that the user does not have permission to view are sorted out (thus not shown in the navbar).


myhpi/core/models.py
```diff
@@ -16,12 +16,14 @@
 class InformationPage(Page):
     body = MarkdownField()
     visible_for = ParentalManyToManyField(Group, related_name="visible_informationpages")
+    is_public = BooleanField()
```
SilvanVerhoeven, line 19: Why don't we add this Boolean to `Page` directly?

myhpi/core/templatetags/core_extras.py
```diff
@@ -2,7 +2,8 @@
...
 register = template.Library()
 
+# logic for this was moved to context.py
 
-@register.filter
-def menu_children(page):
-    return page.get_children().in_menu()
+# @register.filter
+# def menu_children(page):
+#    return page.get_children().in_menu().specific()
```
SilvanVerhoeven, line 9: Comments still in code (remove if code is not needed any more)

myhpi/core/wagtail_hooks.py
```diff
@@ -37,7 +37,8 @@ def check_view_permissions(page, request, serve_args, serve_kwargs):
             target_groups = Group.objects.filter(
                 Q(name=request.user._ip_range_group_name) | Q(id__in=request.user.groups.all())
             )
-        if not any(group in page.visible_for.all() for group in target_groups):
+        matchesGroup = any(group in page.visible_for.all() for group in target_groups)
+        if not (matchesGroup or page.is_public):
```
SilvanVerhoeven, line 41: Renaming `matchesGroup` to `is_matching_group` would be more pythonic, I think

myhpi/core/context.py
```diff
@@ -1,9 +1,36 @@
...
+                    page.menu_children = visible_subpages
+                    page_list.append(page)
+        return page_list
+
+    usergroups = request.user.groups.all()
+    if getattr(request.user, "_ip_range_group_name", None):
+        usergroups = Group.objects.filter(
+            Q(name=request.user._ip_range_group_name) | Q(id__in=request.user.groups.all())
+        )
+
```
SilvanVerhoeven, line 31: May needs to be updated and moved back tot `core_extras.py` after the basic frontend #48 has been merged

myhpi/core/middleware.py
```diff
@@ -20,10 +20,9 @@ def __call__(self, request):
...
-                if address in ip_range:
-                    # user is in this IP range
-                    request.user._ip_range_group_name = group_name
-                    break
+        address = ip_address(request.META.get("REMOTE_ADDR"))
+        for ip_range, group_name in self.ip_ranges.items():
+            if address in ip_range:
+                # user is in this IP range
+                request.user._ip_range_group_name = group_name
+                break
```
SilvanVerhoeven, line 28: Out of curiosity: What is the behaviour of the new code? Do I need to be in the University network even when I am logged in?

.pre-commit-config.yaml
```diff
@@ -16,6 +16,6 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/psf/black
-    rev: 21.4b2
+    rev: 22.3.0
```
SilvanVerhoeven, line 19: hzjuzki

SilvanVerhoeven (changes requested):

.pre-commit-config.yaml
```diff
@@ -16,6 +16,6 @@ repos:
...
     hooks:
```
Paula-Kli, line 20: if you rebase the branch this should be resolved

myhpi/core/models.py
Paula-Kli: Page is something that is taken from wagtail - might this be the point why it is added here?

SilvanVerhoeven: Very well! We could add our own `BasePage` then instead of duplicating code

jeriox: that sounds like a good idea, would also make the permission check easier (currently checks if it is in the list of our page types)

DrEGZo: I'm about to do that but it's hard to migrate when you add a new superclass that wasn't there previously. It will most likely result in a data loss upon migration.

myhpi/core/middleware.py
DrEGZo: Depending on which IP the user has, they are assigned a certain role in addition to the roles that they might already have. If the user is somewhere in the UP network, they might receive a role such as "UP" or any the like. That will affect which of the pages will be visible to the user.
Previously the user would only receive that additional role when being anonymous, which is an issue, because whenever you signed in to your account you were not anonymous anymore, leading to the loss of that role - even if you're still in the same network. And it somehow would be weird if you can view less pages whenever you sign in (kinda counter-intuitive).

myhpi/core/templatetags/core_extras.py
DrEGZo: have been removed

myhpi/core/wagtail_hooks.py
DrEGZo: has been renamed

myhpi/core/context.py
DrEGZo: left unchanged for now, but the menu builder in #48 now uses the menu_children list instead of running another db query

myhpi/core/context.py
```diff
@@ -1,9 +1,36 @@
...
 
 
 def base_context(request):
     root_page = getattr(Site.find_for_request(request), "root_page", None)
+    pages = Page.objects.in_menu().child_of(root_page).specific()
+
+    # filter menu items based on which pages the user can view
+    def can_view(pages, usergroups):
+        page_list = []
+        for page in pages:
```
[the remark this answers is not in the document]
DrEGZo: - 1 page w/ 2 subpages w/ 2 subpages each: 17 Queries, 7ms
- 2 pages w/ 2 subpages each w/ 2 subpages each: 69 Queries, 11ms
- 4 pages w/ 2 subpages each w/ 2 subpages each: 109 Queries, 20ms
- 8 pages w/ 2 subpages each w/ 2 subpages each: 189 Queries, 27ms
Latencies have sometimes high fluctuations, so those values are not reliable. Django already does some query caching but idk if that will be enough in production.

myhpi/core/middleware.py
SilvanVerhoeven: Ohh, I see now: Looks like we can define one role per specific IP range that is then added to the user?
A little comment (or maybe only an extraction method like `assign_group_for_ip_range`) would have made that clearer to me

myhpi/core/context.py
SilvanVerhoeven: So this here pre-processes the available `menu_children`, on which the menu is built then?

myhpi/core/context.py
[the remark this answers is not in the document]
jeriox: that sounds like a lot of queries :/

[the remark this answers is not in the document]
felixrindt: I looked over the code with @jeriox. Right now, this function (which doesn't need to be defined in a function body and could be named `_get_menu_items`) fetches subpages in line 14 each call, and in case of leaf nodes evaluates the `visible_for.all()` queryset for that page. We should avoid evaluating querysets with each call.

We could fetch the whole tree of Pages and Menuitems at once and go through it "in python", without hitting the database using e.g. `get_descendants` on the root_page. I fear it wont be easy to come up with nice code handling the tree, but maybe the graph framework wagtail uses offers some tools for that.

For checking permissions, we first thought we could slap a `prefetch_related("visible_for")` on it, but Menuitems don't have that. Also, right now, there is no single base class defining `visible_for`, so we semantically need to go looking in a different db table for each Page Type. As discussed elsewhere, such a base class is to be introduced, though it might be annoying to have yet another table for that. (btw, why aren't you using djangos object permissions?)

As menu items still wont have `visible_for` you could fetch the "reverse": the list of all pages/page ids that are visible by the user (based on their group memberships). Then you can go walk the tree and check viewing permissions without needing extra queries.

myhpi/core/models.py
DrEGZo: `BasePage` class added.

myhpi/core/context.py
DrEGZo: exactly

myhpi/core/context.py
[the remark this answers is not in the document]
DrEGZo: Base class has been added and process has been reworked. It now produces only 2 queries.

myhpi/core/context.py
```diff
@@ -1,9 +1,65 @@
...
+        path_map[page.path] = page
+        depth_levels.add(page.depth)
+
+    # Iterate over all depth of the tree, going from leaves to root
+    # Mark any parent nodes as non-leaves
+    # If page is allowed to be viewed and (it has viewable children or is a leaf), add it to the parent's viewable children
+    while depth_levels:
+        deepest_level = max(depth_levels)
+        for path in path_map:
+            page = path_map[path]
```
felixrindt, line 48:
```suggestion
        for path, page in path_map.items():
```

myhpi/core/context.py
```diff
@@ -1,9 +1,65 @@
...
+            if page.depth != deepest_level:
+                continue
+            parent_path = page.path[:-4]
+            if parent_path in path_map:
+                path_map[parent_path].is_leaf = False
+                page_allowed_to_be_viewed = page in pages_visible_for_user or page.is_public
+                page_has_children_or_is_leaf = len(page.menu_children) > 0 or page.is_leaf
```
felixrindt, line 55: Could you do just

```suggestion
                page_has_children_or_is_leaf = page.menu_children or page.is_leaf
```

myhpi/core/context.py
```diff
@@ -1,9 +1,65 @@
...
     root_page = getattr(Site.find_for_request(request), "root_page", None)
+    pages = BasePage.objects.in_menu().live()
+    pages_visible_for_user = []
+    path_map = {}
+    depth_levels = set()
+
+    if not root_page:
+        return {"root_page": None, "all_pages": None}
+
+    # Determine all pages the user may view based on his groups
```
frcroth, line 22:
```suggestion
    # Determine all pages the user may view based on their groups
```

myhpi/core/context.py
```diff
@@ -1,9 +1,65 @@
-from wagtail.core.models import Page, Site
+from django.contrib.auth.models import Group
+from django.db.models import Q
+from wagtail.core.models import Site
+
+from .models import BasePage
 
 
 def base_context(request):
+
```
frcroth, line 9:
```suggestion
```

myhpi/core/context.py
```diff
@@ -1,9 +1,65 @@
...
+    # Add additional properties to each page and save in path_map
+    # The path of each page is represented by a string, each level labeled with 4 chars
+    for page in pages:
+        page.menu_children = []
+        page.is_leaf = True
+        path_map[page.path] = page
+        depth_levels.add(page.depth)
+
+    # Iterate over all depth of the tree, going from leaves to root
```
frcroth, line 42:
```suggestion
    # Iterate over entire depth of the tree, going from leaves to root
```
Or what is this comment supposed to say?

frcroth: Also why go from leaves to root and not the other way?

myhpi/settings.py
```diff
@@ -84,6 +85,7 @@
 LOGIN_URL = "login"
 
 MIDDLEWARE = [
+    # "debug_toolbar.middleware.DebugToolbarMiddleware", # Enable this line to get debug toolbar
```
frcroth, line 88: Perhaps add an env variable to automatically add this to middleware?

frcroth (commented):
Also can we clean up the commits (squash some commits together with a rebase). Perhaps also using commit messages like described here: https://cbea.ms/git-commit/ (At least for the titles)

myhpi/static/css/myHPI.css
```diff
@@ -67,10 +67,6 @@ a {
     background-color: var(--bs-body-bg-light);
 }
 
-.navbar-bottom-content .nav-item-container {
-    margin-bottom: 3px;
-}
-
```
SilvanVerhoeven, line 73: Why do you remove this class here (and add the `calc` 3px below)?

myhpi/settings.py
DrEGZo: done

myhpi/core/context.py
DrEGZo: because when none of a page's child pages are visible, the page itself is also not supposed to be visible.
going from root to leaves would require backtracking which I wanted to avoid here.

frcroth (approved):
What could go wrong?

Issue #69, opened 2022-06-05
Expand abbreviations in printed minutes/pages


frcroth:
When printing abbreviations should be expanded, since one can no longer hover over them to get an explanation.

Requires #52.

PR #76, opened 2022-06-07
Add post processing for printing minutes

### LINKED ISSUE #66: Improve links when printing
When a page is printed, underlined text is not useful for links. Instead, links should be written in full (e.g. at the end of the page) and referenced from within the text. Requires #52.

Possible solution: https://thenewcode.com/1186/Better-Links-for-Printed-Web-Pages-with-JavaScript

frcroth:
Closes #66, #69 and implements better handling for links (numbers and a list at the bottom of the page) and abbreviations (expands them).

There is still a remaining problem. When only one link is present, the rendering breaks on Firefox for unknown reasons. It works with multiple links on Firefox, it works with one or more links on Chromium, I have only observed this one weird case. We need to decide if that is ok, since it is only one edge case, if we should display a warning in that one case or if that needs to be fixed, which I have no idea how to do.

Firefox correct:
image

Chromium correct:
image

Firefox incorrect with one link:
image

Chromium correct with one link:
image

### DIFF
--- a/myhpi/core/templates/core/minutes.html
+++ b/myhpi/core/templates/core/minutes.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load static %}
 {% load wagtailmarkdown %}
 
 {% block content %}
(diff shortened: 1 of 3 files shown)


jeriox (approved):
I'm fine with merging it as is, we can still create an issue afterwards if it bothers us too much.

Paula-Kli:
Cannot reproduce

Issue #92, opened 2022-06-11
Minutes list shows unpublished entries


frcroth:
Minutes list does not check if minutes are published or not.

jeriox:
Maybe we should also filter the displayed minutes by whether the current user is allowed to view them

frcroth:
Yes, of course, though this would need to wait for #57, when utilities are added to determine which pages can be viewed. I would not implement this here because of code duplication.

jeriox:
If the user has edit permissions/can add minutes, the list should also display unpublished minutes

frcroth:
@Paula-Kli Sorry, I already handled this in #93, it seemed natural to solve these two issues at once.

PR #93, opened 2022-06-11
Add minutes navigation

### LINKED ISSUE #43: Minutes navigation
Buttons for navigating to next/previous minutes document on minutes view page

frcroth:
Closes #43, #92, #98.

---OUTDATED INFO BELOW---

Closes #43. However, while it does check that we can not navigate to unpublished pages, it does not check for permissions to view the page (Leading to information leakage and user frustration). Checking permissions of minutes is also a topic for #92. So possibly the correct order is merging #57 with good utilities to check if a collection can be viewed (a filter on a page queryset would be best), and then #92 and this.

image


myhpi/core/models.py
```diff
@@ -107,6 +107,21 @@ class Minutes(Page):
     parent_page_types = ["MinutesList"]
     subpage_types = []
 
+    @property
+    def next_minutes(self):
+        siblings = self.get_next_siblings().live()
+        # TODO: When #57 is merged, iterate through pages and check if user can view them
+        if siblings:
+            return siblings[0]
```
jeriox, line 115: what is the order here? will the first item in the list always be the minutes that we are looking for? or is this influenced by time of last edit or similar?

frcroth: I think it is creation time (the order of pages in the tree)

jeriox: This may sometimes be the wrong order, e.g. because minutes of GF-Treffen are written by hand and digitized afterwards. There may have been another meeting in between.

myhpi/core/models.py
```diff
@@ -79,12 +88,20 @@ class MinutesList(BasePage):
...
-        context = super().get_context(request, *args, **kwargs)
+    def get_visible_minutes(self, request):
         minutes_ids = self.get_children().exact_type(Minutes).values_list("id", flat=True)
+        user_groups = get_user_groups(request.user)
         minutes_list = (
-            Minutes.objects.filter(id__in=minutes_ids, group=self.group).order_by("date").reverse()
+            Minutes.objects.filter(id__in=minutes_ids, group=self.group)
+            .filter(Q(visible_for__in=user_groups) | Q(is_public=True))
+            .order_by("date")
+            .reverse()
```
jeriox, line 98: not sure what reverse does, but it is probably more efficient to order by `-date`

myhpi/core/models.py
```diff
@@ -79,12 +88,20 @@ class MinutesList(BasePage):
...
     subpage_types = ["Minutes"]
 
-    def get_context(self, request, *args, **kwargs):
-        context = super().get_context(request, *args, **kwargs)
+    def get_visible_minutes(self, request):
         minutes_ids = self.get_children().exact_type(Minutes).values_list("id", flat=True)
+        user_groups = get_user_groups(request.user)
         minutes_list = (
-            Minutes.objects.filter(id__in=minutes_ids, group=self.group).order_by("date").reverse()
+            Minutes.objects.filter(id__in=minutes_ids, group=self.group)
```
jeriox, line 95: we wanted to remove the group field from minutes (#98), you could also include that in this PR :)

jeriox (changes requested):

myhpi/core/models.py
```diff
@@ -29,6 +29,15 @@ class BasePage(Page):
     ]
 
 
+def get_user_groups(user):
```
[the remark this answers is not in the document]
frcroth: Do we have access to the user?

myhpi/core/models.py
frcroth: I don't know about efficient, but possibly more legible

myhpi/core/models.py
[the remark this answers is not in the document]
jeriox: ah well, I was in the mindset for my other project where we use a custom user model. as we don't do this here i'm fine with the separate method. but maybe move it to something like utils.py as it is not a model?

myhpi/core/models.py
frcroth: Done

myhpi/core/models.py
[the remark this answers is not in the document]
frcroth: Done

myhpi/core/models.py
```diff
@@ -92,10 +84,9 @@ def get_visible_minutes(self, request):
         minutes_ids = self.get_children().exact_type(Minutes).values_list("id", flat=True)
         user_groups = get_user_groups(request.user)
         minutes_list = (
-            Minutes.objects.filter(id__in=minutes_ids, group=self.group)
+            Minutes.objects.filter(id__in=minutes_ids)
             .filter(Q(visible_for__in=user_groups) | Q(is_public=True))
```
jeriox, line 88: wouldn't it be possible to apply this filter to self.get_children().exact_type(Minutes) instead?

jeriox (approved):

myhpi/core/models.py
frcroth: Don't know why this is this way

frcroth: But it doesn't work the other way, so I'm just going to merge

PR #94, opened 2022-06-11
Add table for minute list

### LINKED ISSUE #87: Redesign minutes list
The minutes list is currently ugly with only bullet points. We'd like to have a column-based design similar to 1327 which also shows the labels of the minutes and visibility in a nice way
grafik

frcroth:
Closes #87.

Remaining issues:
- ~~Labels do not have an associated color, so the secondary color is used~~ -> moved to #95 
- ~~Font awesome icons are not yet available~~ -> Not entirely relevant for this PR since currently only one icon is needed
- ~~It is not clear what should be shown in the permissions field~~ -> See discussion on #87
- ~~There is no separation by years, since it does not seem immediately important~~ Done.

image

### DIFF
--- a/myhpi/core/models.py
+++ b/myhpi/core/models.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from django import forms
 from django.contrib.auth.models import Group, User
 from django.db import models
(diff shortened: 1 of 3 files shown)


jeriox:
> Remaining issues:
> 
>     * Labels do not have an associated color, so the secondary color is used
do you want to add a color in this PR? move to separate issue? do we need color at all? They made a quick distinction possible between the different types, but I guess it would be fine without colors as there are only a few minutes with labels
>     * Font awesome icons are not yet available
@SilvanVerhoeven did you leave them out when installing bootstrap on purpose or didn't you just see the need?
>     * It is not clear what should be shown in the permissions field
I made the same proposal on the issue as you just implemented, so I'm fine with that
>     * There is no separation by years, since it does not seem immediately important
that made it easier to find a date range, e.g. if you wanted to look up something in a protocol that was taken in june 2021. another reason was that the list was so long and to provide some visual separation. we could also paginate the view based on the years, that could also speed up the page load

jeriox (approved):
lgtm!

SilvanVerhoeven:
> 
> > ```
> > * Font awesome icons are not yet available
> > ```
> 
> @SilvanVerhoeven did you leave them out when installing bootstrap on purpose or didn't you just see the need?

Should they come with Bootstrap? I reckon to have installed the full bootstrap version and added bootstrap icons later. Haven't thought about Font Awesome.

SilvanVerhoeven:
> 
> > ```
> > * Font awesome icons are not yet available
> > ```
> 
> @SilvanVerhoeven did you leave them out when installing bootstrap on purpose or didn't you just see the need?

@jeriox Should they come with Bootstrap? I reckon to have installed the full bootstrap version and added bootstrap icons later. Haven't thought about Font Awesome.

jeriox:
> 

no, they need to be installed separately as boostrap now has its own iconset, which can be installed as an addon

PR #111, opened 2022-06-28
Explicitly assign reviewer for dependabot PRs

frcroth:
Since #106 didn't work with the reviewer team, as seen in #110, this should work (at least it has before).


jeriox (approved):
would prefer not having individual user names in the code, but if the team stuff doesn't work....

frcroth:
> I don't think that this is a nice solution too, but if it works woman_shrugging do you want to assign anybody else?

Well let's see if it work 😅 At least that's the solution we used in our BP.

PR #113, opened 2022-06-30
Remove assignment for dependabot PRs

frcroth:
Because we can't have nice things.
image


Paula-Kli (approved):
Oh that is sad... well I will try to quickly review them anyways

PR #137, opened 2022-07-24
Add search

### LINKED ISSUE #7: Set up search
- [ ] index model fields (https://docs.wagtail.org/en/stable/topics/search/indexing.html#indexing-extra-fields)
- [ ] provide search box and result view
- [x] decide on search backend

frcroth:
Closes #7 

The search page can be found at /search, we might want to discuss adding the search field to the base page, but that is a design question.

The filtering creates a bunch of warnings which is a little annoying, maybe they should be suppressed.


myhpi/core/models.py
```diff
@@ -29,6 +30,11 @@ class BasePage(Page):
         FieldPanel("is_public", widget=forms.CheckboxInput),
         FieldPanel("visible_for", widget=forms.CheckboxSelectMultiple),
     ]
+    # FilterFields required for restricting search results
+    search_fields = Page.search_fields + [
+        index.FilterField("group_id"),
```
jeriox, line 35: do we really want to search for the group id?

jeriox (commented):
just the first thought, want to check it out and have a look in the next few days

myhpi/core/models.py
frcroth: No we don't. This field is required for permission checks (so we don't output search results the user can't visit)

jeriox: I don't understand why this is working, as searching on filtered QuerySets should not be possible as of https://github.com/wagtail/wagtail/issues/6616. As ist works right now, I would be fine with accepting that even if it generates warnings

jeriox (approved):

frcroth:
Merging now. Opens up #140 and #141

PR #139, opened 2022-07-25
Remove bootstrap from repository

### LINKED ISSUE #67: Bootstrap in the repo
I think it is not optimal that bootstrap is included in the repo (https://github.com/fsr-de/myHPI/tree/main/myhpi/static/scss/bootstrap), since it now makes up the majority of the codebase. 

It would be better to make this a dependency or automate the installation in some other way.

image

frcroth:
Closes #67


jeriox (approved):
I think this will break if we want to build wheels from the repo with `poetry build` at some point. But as we are currently depending on docker images, it should be fine if we include the installation there

Issue #88, opened 2022-06-10
Remove dummy page from migrations


frcroth:
Create a good root page

PR #143, opened 2022-07-31
Add tests

### LINKED ISSUE #140: Add test for search
Since search is somewhat hacky right now, create a test.

frcroth:
Closes #140 and adds some wagtail tests.

~~So it seems like the CI is still broken, since the database it uses does not have the ~~

See comment below for current problem
---

~~Edit: Blocked by #146~~

~~Also, while working on this I had constant issues with static files. It seems like I can't compile them right now, because some source map is missing for bootstrap js. Looking back at CI logs, this problem has been there since the introduction of bootstrap to the repo. It is a problem since the tests fail when I do not include them (locally they fail because of that, here they fail beforehand).
When I add that manually, the compilation fails because of bootstrap scss, where in _root.scss the $theme_colors variable is not found. Really weird stuff. I had to remove all static files to get it running locally again (but I didn't commit it, because I don't see what the connection with tests is)~~


jeriox (approved):
I am used to working with pytest instead of the django built-in test engine, but as you already wrote all those tests i think I'm fine with sticking with that

frcroth:
So I merged the migrations, since they were only valid on sqlite. It seems like the search works on sqlite and postgres, but not on mysql. How to go from here?
1. Ignore these tests while testing with mysql
2. Removing mysql from tested dbs
3. Removing the tests
4. Removing the search functionality

@jeriox @DrEGZo happy to hear your ideas

DrEGZo:
#88 would then depend on this.

Since the migrations needed restructuring anyways, I'm fine w/ merging them all in one. That seems like a straight-forward solution... I don't believe we need the history aspect of the previous migrations that badly.

frcroth:
> #88 would then depend on this.
> 
> Since the migrations needed restructuring anyways, I'm fine w/ merging them all in one. That seems like a straight-forward solution... I don't believe we need the history aspect of the previous migrations that badly.

Merging the migrations... But what about the mysql issue?

jeriox:
I'm fine with dropping mysql. We can use postgres in production and are still left with sqlite for local development, so I don't see why we should stick to mysql

PR #144, opened 2022-07-31
Improve bootstrap installation

frcroth:
Since bootstrap was originally added to the repository, we were missing a file (bootstrap.min.js.map). This adds the file to the automatic install process.

Also improves the automatic install process so you can update your local bootstrap.


frcroth:
Also on this branch I try to fix the static compilation, which is currently broken on my machine, I don't know why.
image

jeriox (approved):

PR #168, opened 2022-09-04
Make CI green: Use new method of installing poetry

frcroth:


jeriox (approved):
poetry also recommends pinning the poetry version (https://python-poetry.org/docs/#ci-recommendations), maybe we should consider that as well. as it would have prevented breaking your pipeline on the other PR in the first place

PR #177, opened 2022-09-15
Guests in minutes

### LINKED ISSUE #175: Guests for minutes
@jeriox Any reason why this is not yet included?

I would not be happy to remove guest information when migrating minutes from 1327

frcroth:
Adding guests is not the most straight-forward way in the interface, since you need to write a JSON. The alternative would have been to create a new guests model, but I think this is a fine database representation (better than in 1327) and the edit widget can be changed at some point.

Also to make the migration smoother I made moderator and author of minutes nullable. 

Closes #175, #178


myhpi/core/models.py
```diff
@@ -152,11 +152,16 @@ class UserSelectWidget(s2forms.ModelSelect2MultipleWidget):
...
+    moderator = ForeignKey(
+        User, blank=True, null=True, on_delete=models.PROTECT, related_name="moderator"
+    )
+    author = ForeignKey(
+        User, blank=True, null=True, on_delete=models.PROTECT, related_name="author"
+    )
     participants = ParentalManyToManyField(User, related_name="minutes")
     labels = ClusterTaggableManager(through=TaggedMinutes, blank=True)
     text = CustomMarkdownField()
+    guests = models.JSONField(blank=True, default=[])
```
jeriox, line 164: JSON feels a bit hacky to me at this point. Maybe we want to approach this similar to abbreviation explanations?

frcroth: I want to be able to add guests on the same page and not need to create a guest first in some other menu and then select that on the minutes page

frcroth: We want to store a list of names, with no additional information. JSON is a type used to store a list of strings, so I think in the db it is fine.

jeriox: all right, makes sense

myhpi/tenca_django/migrations/0002_alter_hashentry_options_delete_legacyadminurl.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.0.7 on 2022-09-15 19:17
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("tenca_django", "0001_initial"),
```
jeriox, line 9: not sure if you actually want that in this PR

myhpi/polls/migrations/0009_alter_poll_description.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations
+
+import myhpi.core.markdown.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("polls", "0008_polllist"),
```
jeriox, line 11: not sure if you actually want that in this PR

jeriox (approved):

myhpi/tenca_django/migrations/0002_alter_hashentry_options_delete_legacyadminurl.py
frcroth: It has to be in some PR, since it is created when running makemigrations

jeriox: you _could_ move it to a separate PR so that one is able to find it afterwards, as no one would assume that it is hidden in this PR. but I'm also fine with merging as-is

frcroth: I think that is not a problem since I made a specific commit for that change and don't squash

PR #188, opened 2022-09-23
Add redirect menu item

### LINKED ISSUE #180: Create lists item in menu
The lists dashboard needs a menu entry. This has to implemented when rendering the menu, since it is not a page.

frcroth:
Closes #180 

I considered adding it automatically to the menu, but I think this is the better way to do it.


myhpi/core/models.py
```diff
@@ -215,3 +215,12 @@ class AbbreviationExplanation(Model):
 
     def __str__(self):
         return self.abbreviation
+
+
+class MailingListsMenuItem(BasePage):
```
jeriox, line 220: this is quite specific. What do you think about adding a RedirectMenuItem instead on which the user can set a target? This would also allow for external links to be set in the menu, sounds like something one could probably need in the future anyways

frcroth: Good idea!

jeriox (approved):

PR #198, opened 2022-10-09
Add Footer

SilvanVerhoeven:
Adds a basic footer

grafik

Notice the management command `python manage.py initializesite` to automatically add the footer.

## Open Questions

1. What exactly do we want to link in the Repo (e.g. should the HPI Campus Discord and student-telegram group be linked)?
2. Follow up: If they should be linked, how can we display them only to logged in users? And how should we define the secret links best to not publish them on GitHub?
3. How to link the English/German version of myHPI in the `Sprachen` column?
4. How to link to pages? Currently, information pages like "Impressum" are descendants of "Footer" to be displayed in the footer. Idiomatic, but pollutes the URL (i.e. `myhpi.de/footer/rechtliches/impressum` vs. `myhpi.de/impressum`).
5. Is there a way to programmatically delete pages? If I delete an earlier footer in `python manage.py initializesite` with `FooterBase.objects.all().delete()` to avoid getting multiple footers when running the command again, I cannot add new children to root until I ran `python manage.py fixtree`.


frcroth (changes requested):
Visually, I think the footer looks good. However, I think the current approach has some downsides, for example that the data regulation page is a child of the footer might not make too much sense. I think the better way of doing things would be to have one new page type, a FooterPage, which is rendered in the way that is shown by you in your screenshot. This FooterPage can have a table that is layouted in a special "footery" way and includes just lists of links that may be internal links to other page (e.g. data regulations, which would be a subpage of the root page that is not displayed in menus) and external links, like social media.

SilvanVerhoeven:
@jeriox Mentioned snippets as a possible solution for this issue today. Will look into this

SilvanVerhoeven:
I have re-implemented the footer as a snippet.
There are some things I am not totally happy with, but I do not want to tackle them now:

1. The external link tagging currently works via checking the base domain of a link - What do links to another wagtail page look like? (myHPI used a special format for internal links)
2. The snippet does not support hiding links from the general public (e.g. for Discord and Telegram). Would need a special markdown syntax
3. Make columns 2 to 4 in footer optional (currently all columns require content)
4. Make number of columns dynamic via Wagtail GUI
5. Adding the external link tagger to all content in myHPI
6. Open external links in a new tab by default

frcroth:
You can create internal links with `[document link](doc:3) ` (with 3 being the document id)

frcroth (approved):
Looking good! I think for the moment we only need information in the footer that is available for everyone so that we don't need to worry about public/internal footers. Please make the tests green. Opening external pages in a new tab would be pretty easy to do with your external filter.

myhpi/core/management/commands/initializesite.py
```diff
@@ -0,0 +1,18 @@
...
+from myhpi.core.models import Footer
+
+
+class Command(BaseCommand):
+    help = "Initializes the website (e.g. creates the default footer)"
+
+    def handle(self, *args, **options):
+        self._add_footer()
+
+    def _add_footer(self):
```
jeriox, line 12: we already add the root page in our own migration. I would suggest adding the footer there as well. We can leave this for now and move it to a separate issue

jeriox (approved):

PR #199, opened 2022-10-09
Add search button

SilvanVerhoeven:
- Adds a search button to the navbar
- Improves design of search page

grafik
grafik
grafik
grafik

### Open Questions

1. How to internationalise the search page?
2. (Under which circumstances) Will the `result.search_description` in `search_result.html` be set?
3. Search is a separate module: Is there a way (and is it preferred) to make the `search.js` and the search button popup in `base.html` part of the module?

### DIFF
--- a/myhpi/search/templates/search/search.html
+++ b/myhpi/search/templates/search/search.html
@@ -3,30 +3,43 @@
 
 {% block body_class %}template-searchresults{% endblock %}
 
-{% block title %}Search{% endblock %}
+{% block title %}Suche{% endblock %}
 
 {% block content %}
-    <h1>Search</h1>
+<h1 class="page-title">
+    <span class="underline">
+        Suche
+    </span>
+</h1>
 
-    {% include "search/search_field.html" %}
-    <p>
-    {% if search_results %}
-            {% for result in search_results %}
-                    <h4><a href="{% pageurl result %}">{{ result }}</a></h4>
-                    {% if result.search_description %}
-                        {{ result.search_description }}
-                    {% endif %}
-            {% endfor %}
+{% include "search/search_field.html" %}
 
-        {% if search_results.has_previous %}
-            <a href="{% url 'search' %}?query={{ search_query|urlencode }}&amp;page={{ search_results.previous_page_number }}">Previous</a>
-        {% endif %}
+{% if search_results %}
 
-        {% if search_results.has_next %}
-            <a href="{% url 'search' %}?query={{ search_query|urlencode }}&amp;page={{ search_results.next_page_number }}">Next</a>
-        {% endif %}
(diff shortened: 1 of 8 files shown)


frcroth (changes requested):
I like the hotkey and I wish for more! The integration of the search bar is nicely done.

I think you don't have to worry about localization in this PR, since we did not previously.  
I still have some problems however.

When I search for a term with many occurences, it only shows 10 results and displays that only 10 results have been found (which is incorrect):
image
Also the layouting of the "Next" does not seem very nice.

When I search for a rare term, it looks as described.

No results page looks good.

Minutes currently don't seem to show a preview, which is not perfect.

Possibly it would be more idiomatic to include search items in the search module, but I don't really care about it.

myhpi/search/templates/search/search.html
```diff
@@ -3,30 +3,43 @@
...
+
+{% include "search/search_field.html" %}
+
+{% if search_results %}
+
+<i class="d-block mb-3 ps-1">
+    {% if search_results|length == 1 %}
+    Ein Ergebnis gefunden.
+    {% else %}
+    {{ search_results|length }} Ergebnisse gefunden.
```
jeriox, line 23: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#pluralize is your friend :)

myhpi/search/templates/search/search_result.html
```diff
@@ -0,0 +1,15 @@
...
+<a href="{% pageurl result %}">
+    <div class="search-result">
+        <h5>{{ result.title }}</h5>
+        <span class="text-dark">
+            {% if result.search_description %}
+            {{ result.search_description }}
+            {% else %}
+            {{ result.specific.body|truncatewords:20 }}
+            {% endif %}
+        </span>
```
frcroth, line 13: This does not work for minutes as the field on minutes is called "text" (See https://github.com/fsr-de/myHPI/issues/206).

```suggestion
        <span class="text-dark">
            {% if result.search_description %}
                {{ result.search_description }}
            {% else %}
                {% if result.specific.text %}
                    {{ result.specific.text|truncatewords:20 }}
                {% else %}
                    {{ result.specific.body|truncatewords:20 }}
                {% endif %}
            {% endif %}
        </span>
```

myhpi/search/templates/search/search.html
SilvanVerhoeven: Good to know! But in this case here, if I want to print "Ein Ergebnis" instead of "1 Ergebnis", this still is the cleaner version, isn't it?

frcroth (approved):
Nice! The only gripe I have is that the search result pagination should have some padding to the bottom of the white area 
image

PR #216, opened 2022-10-31
fix: Make menu scrollable on Desktop

SilvanVerhoeven:
Should be working now (testing it in depth is cumbersome for me, as I'd have to create loads of pages manually/duplicate nodes in DOM)


frcroth (commented):
Does not work for me because navbar-bottom has the overflow-hidden class

SilvanVerhoeven:
Now?

frcroth (approved):
Seems to work now

PR #218, opened 2022-11-02
add page type to allow for voucher obtaining

jeriox:
As a result of the halloween party, we want to improve the flow for ticket sales. The proposed flow realised by this app looks as follows:

The user refers to myhpi.de/weihnachten and is greeted with some information about the event. The page determines whether the user is an angel or not. The user is informed about the consequences and offered a button to create vouchers (obtained from pretix). With these vouchers, the user is able to buy a ticket in the ticket shop. If the user is an angel, an angel voucher is issued. 

grafik

ToDo
- [x] API connection to pretix -> obtain vouchers from pretix
- [x] add link to redeem code directly
- [ ] connection to engelsytem
- [ ] display warning for non-angels that they should sign up beforehand
- [ ] make pretty


jeriox:
@SilvanVerhoeven any thoughts on how to display the vouchers in a nice way? the link refers to the ticket shop and auto-applies the voucher

SilvanVerhoeven:
To get the full picture: This workflow should replace the manual ticket type selection in Pretix at the beginning?
What does the voucher do on the Pretix side and how is it transfered to Pretix?

jeriox:
Suggestion by Aaron:
grafik
grafik

jeriox:
> To get the full picture: This workflow should replace the manual ticket type selection in Pretix at the beginning? What does the voucher do on the Pretix side and how is it transfered to Pretix?

The voucher is requested in pretix via the API. It allows the user to buy a corresponding ticket (angel or regular) which will only be possible with such a voucher in the beginning

PR #219, opened 2022-11-03
add Markdown preprocessor to decrease heading level

### LINKED ISSUE #45: Remap heading levels from markdown to HTML
Currently, `#` is mapped to `<h1>` and so on. As the page title is rendered as `<h1>` as well, we should map `#` to `<h2>`.
Preferrably, this should include the new structure from #44

jeriox:
closes #45 

There is one small issue left with the current solution: a heading of level 6 is still matched by the regex, as there are 5 hashtags in it. This means that one is left with a result of `<h6># Text</h6>`. The smoother solution would be to not replace the line at all, but I can't come up with a regex that does so. As nobody actually uses h6 anyways, I think thats not such a big issue

### DIFF
--- a/myhpi/core/markdown/extensions.py
+++ b/myhpi/core/markdown/extensions.py
@@ -133,6 +133,17 @@ class EnterLeavePreprocessor(MinutesBasePreprocessor):
         return self.enter_or_leavify(match, _("leaves"))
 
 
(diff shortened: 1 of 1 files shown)


frcroth (approved):

jeriox:
@frcroth thx for the review. Do you have an opinion on the problem mentioned in the PR description? if not, feel free to merge

PR #221, opened 2022-11-06
move footer creation to migration

### LINKED ISSUE #211: Move footer creation into root page creation migration
The footer is currently created via commands (`commands/initializesite.py`).

We already have a migration to add the root page. I would suggest adding the footer there as well. We can leave this for now and move it to a separate issue

_Originally posted by @jeriox in https://github.com/fsr-de/myHPI/pull/198#discussion_r1004388028_

jeriox:
closes #211


frcroth (approved):

frcroth:
Just linting needed

SilvanVerhoeven:
Ahh sorry that I forgot to close the issue.  
Why did you work the migration into previous migrations?

SilvanVerhoeven:
Ahh sorry that I forgot to close the issue.  
@jeriox Why did you work the migration into previous migrations?

jeriox:
as the code cannot work with no footer being available, I wanted to make sure that there cannot be any instance with a database that is in a state without a footer

Issue #179, opened 2022-09-16
Logo is way too big


frcroth:
![image](https://user-images.githubusercontent.com/6863832/190635926-ceae49e2-1f1d-4b8b-8d71-091d750a4531.png)
Some content and the menu is no longer visible.

```
Error: <svg> attribute width: Expected length, "auto".
localhost/:108 Error: <svg> attribute width: Expected length, "auto".
localhost/:153 Error: <svg> attribute width: Expected length, "auto".
localhost/:186 Error: <svg> attribute width: Expected length, "auto".
localhost/:419 Error: <svg> attribute width: Expected length, "auto".
localhost/:460 Error: <svg> attribute width: Expected length, "auto".
localhost/:793 Error: <svg> attribute width: Expected length, "auto".
localhost/:926 Error: <svg> attribute width: Expected length, "auto".
localhost/:995 Error: <svg> attribute width: Expected length, "auto".
localhost/:1208 Error: <svg> attribute width: Expected length, "auto".
localhost/:1297 Error: <svg> attribute width: Expected length, "auto".
localhost/:1330 Error: <svg> attribute width: Expected length, "auto".
localhost/:1447 Error: <svg> attribute width: Expected length, "auto".
localhost/:1564 Error: <svg> attribute width: Expected length, "auto".
localhost/:1681 Error: <svg> attribute width: Expected length, "auto".
localhost/:1758 Error: <svg> attribute width: Expected length, "auto".
localhost/:1803 Error: <svg> attribute width: Expected length, "auto".
localhost/:1912 Error: <svg> attribute width: Expected length, "auto".
localhost/:2085 Error: <svg> attribute width: Expected length, "auto".
localhost/:2146 Error: <svg> attribute width: Expected length, "auto".
localhost/:2227 Error: <svg> attribute width: Expected length, "auto".
localhost/:2288 Error: <svg> attribute width: Expected length, "auto".
localhost/:2321 Error: <svg> attribute width: Expected length, "auto".
localhost/:2406 Error: <svg> attribute width: Expected length, "auto".
localhost/:2483 Error: <svg> attribute width: Expected length, "auto".
localhost/:2552 Error: <svg> attribute width: Expected length, "auto".
localhost/:2605 Error: <svg> attribute width: Expected length, "auto".
localhost/:2682 Error: <svg> attribute width: Expected length, "auto".
localhost/:2747 Error: <svg> attribute width: Expected length, "auto".
[...]

PR #284, opened 2023-02-23
Minor Improvements and Fixes

### LINKED ISSUE #276: Hide forbidden pages in menu
Pages not accessible for the current user are displayed in the menu. No no no!

### LINKED ISSUE #277: Equal font size in bread crumbs
The font size of the current site's breadcrumb is larger than the other breadcrumbs

### LINKED ISSUE #278: Remove visual right offset of most right button in header
The button most right in the header seems to be offset from the right. It is not, but the padding around the icon (so users can properly hit the button) makes it look so.
To fix this, move a most right button more to the right, so it visually appears to align properly with the page content

Current situation (with the menu button as an example):
grafik

Imagined solution:
grafik

### LINKED ISSUE #279: Hide login label in mobile layout
On some mobile devices, the "Anmelden" login label and login icon do not fit into one row, leading to a nasty looking line break. Suggested solution: remove the label.
grafik

### LINKED ISSUE #280: Display menu button in mobile layout
Duh.

SilvanVerhoeven:
Close #280 close #279 close #278 close #277 close #276

### DIFF
--- a/myhpi/core/context.py
+++ b/myhpi/core/context.py
@@ -1,4 +1,5 @@
-from wagtail.core.models import Page, Site
+from django.db.models import Q
+from wagtail.core.models import Site
 
 from .models import BasePage
 from .utils import get_user_groups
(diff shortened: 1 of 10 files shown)


frcroth (commented):
The style changes are all well and good, but the new menu generation causes problems. When I'm in multiple groups that are allowed to view a menu item, that menu item is shown multiple times.

SilvanVerhoeven:
@frcroth Currently, pages that are not live yet are generally not displayed in the menu. This also means that an unpublished private minute for e.g. the student representative group is not displayed in the menu.

What is the expected behaviour?

jeriox:
when trying this out I am encountering the big logo that has already occured in the past. I removed bootstrap and the collected static files, but it persists

SilvanVerhoeven:
> when trying this out I am encountering the big logo that has already occured in the past. I removed bootstrap and the collected static files, but it persists
> grafik
> 

That's odd. I don't think I changed anything at the logo files. Can anyone reproduce this?

frcroth:
> @frcroth Currently, pages that are not live yet are generally not displayed in the menu. This also means that an unpublished private minute for e.g. the student representative group is not displayed in the menu.
> 
> What is the expected behaviour?

Why would unpublished minutes be displayed in the menu? Unpublished minutes are currently displayed in the minutes list, but where would we want to have minutes in a menu?

frcroth:
> when trying this out I am encountering the big logo that has already occured in the past. I removed bootstrap and the collected static files, but it persists

Having the same issue. Deleted static files like in #179 with no change.

SilvanVerhoeven:
> > @frcroth Currently, pages that are not live yet are generally not displayed in the menu. This also means that an unpublished private minute for e.g. the student representative group is not displayed in the menu.
> > What is the expected behaviour?
> 
> Why would unpublished minutes be displayed in the menu? Unpublished minutes are currently displayed in the minutes list, but where would we want to have minutes in a menu?

They could be if someone chooses to select "display in menu" in the "publication" tab of a minute. Since minutes are BasePages they could theoretically be displayed in the menu

However, I now realize this is a more fundamental question: Should unpublished pages (drafts) be displayed in the menu or not? Right now, only published pages with "display in menu" are displayed in the menu of users with the appropriate viewing rights.

Any meaning @jeriox @frcroth

SilvanVerhoeven:
> when trying this out I am encountering the big logo that has already occured in the past. I removed bootstrap and the collected static files, but it persists grafik

Should be fixed now, please recheck

frcroth:
This is a really hard question to answer, since we do not know how the page is used. Through the display in menu feature we basically already have some kind of draft status and it is not clear if the draft status is of any benefit here.

My anecdotal user testing has also shown that people found it weird that "save as draft" is the default option to save the page and "Publish page" does not actually publish the page since you need to select visibility options. So in the future it might make sense to remove the draft feature all together - or reevaluate how it interacts with other features.

Since we will not do that now, I think the correct way of going forward with drafts would be using drafts as intended by wagtail. In that way I don't think they should be displayed in menus.

jeriox:
> > > @frcroth Currently, pages that are not live yet are generally not displayed in the menu. This also means that an unpublished private minute for e.g. the student representative group is not displayed in the menu.
> > > What is the expected behaviour?
> > 
> > 
> > Why would unpublished minutes be displayed in the menu? Unpublished minutes are currently displayed in the minutes list, but where would we want to have minutes in a menu?
> 
> They could be if someone chooses to select "display in menu" in the "publication" tab of a minute. Since minutes are BasePages they could theoretically be displayed in the menu
> 
> However, I now realize this is a more fundamental question: Should unpublished pages (drafts) be displayed in the menu or not? Right now, only published pages with "display in menu" are displayed in the menu of users with the appropriate viewing rights.
> 
> Any meaning @jeriox @frcroth

I think we should keep it that way, I can't think of a scenario where having an unpublished minutes in the menu would be a good idea.

SilvanVerhoeven:
Awesome, then I think this is ready to merge

Looking forward to having "Publish" the default ^^

jeriox (approved):

PR #297, opened 2023-03-04
Update settings for mail and logging

### LINKED ISSUE #142: When an exception triggers, send a mail to designated admins

frcroth:
Fixes #142 

Copied settings from https://github.com/ephios-dev/ephios/blob/main/ephios/settings.py


.env.example
```diff
@@ -8,7 +8,7 @@ MEDIA_URL=/media/
 MEDIA_ROOT=data/media/
 INTERNAL_IPS=["127.0.0.1","localhost"]
 SITE_URL=http://localhost:8000
-EMAIL_URL=dummymail://
+EMAIL_URL=smtp://
```
jeriox, line 11: does this throw errors when used as is? maybe we should keep dummymail in the example to allow for local development without changes

jeriox (approved):

.env.example
frcroth: Would be fine with that. As is no errors are thrown because emails are only sent if Debug=FALSE

jeriox: ah, I see. Would still vote for the dummymail default as we will integrate other features that send mails (minutes reminder etc.) which would also send with DEBUG=True

PR #327, opened 2023-04-09
Use env file for anonymous IP range and support multiple admins

frcroth:


myhpi/settings.py
```diff
@@ -240,7 +240,9 @@
 # e.g. in notification emails. Don't include '/admin' or a trailing slash
 WAGTAILADMIN_BASE_URL = SITE_URL
 
-ANONYMOUS_IP_RANGE_GROUPS = {"127.0.0.1": "Moderators"}
+ANONYMOUS_IP_RANGE_GROUPS = {
+    env.str("ANONYMOUS_IP_RANGE"): env.str("ANONYMOUSE_IP_RANGE_GROUP_NAME")
```
jeriox, line 244: typo

jeriox (approved):

myhpi/settings.py
frcroth: ah forgot to commit the fix

PR #376, opened 2023-08-09
Add abbreviation explanations on mobile

### LINKED ISSUE #363: View abbreviations on mobile
Currently, you can only view the explanation for an abbreviation (e.g., "PDF - Preis des Fachschaftsrats) on the desktop by hovering the element. It would be nice if on mobile, I could tap the abbreviation to get a small pop-up.

frcroth:
Closes #363

Nice to have todo: Write abbreviations tabindex in markdown compiler, not in js


abc013:
Abbreviation now uses a filter instead of js.

myhpi/core/templatetags/core_extras.py
```diff
@@ -53,6 +53,16 @@ def tag_external_links(content):
     return template.render(Context())
 
 
+@register.filter(name="touchify_abbrevations")
+def touchify_abbrevations(content):
+    """Takes the content of a website and inserts external link icons after every external link."""
+    external_links = re.finditer("<abbr[^>]*>[^<]*", content)
```
jeriox, line 59: these are not external links

myhpi/core/templatetags/core_extras.py
```diff
@@ -53,6 +53,16 @@ def tag_external_links(content):
...
+    for link in reversed(list(external_links)):
```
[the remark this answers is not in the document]
frcroth: Because of the way this is done with indices, otherwise indices would move.

myhpi/core/templatetags/core_extras.py
abc013: Argh, silly me didn't rename anything. Sorry.

jeriox (approved):

PR #377, opened 2023-08-09
use proper image rendering instead of preview

### LINKED ISSUE #5: Image Chooser shouldn't display preview
Currently if you insert an image, the image is displayed in a preview size.
This should be changed to display a higher resolution (800px is generated from wagtail).

Use a regex to find the higher resolution path from the preview path

jeriox:
closes #5


frcroth (approved):

frcroth:
Followup: test would be nice here

PR #380, opened 2023-08-15
use localized page tree for menu, add language chooser to footer

jeriox:
@frcroth this produces a menu with the english titles with my local dataset. Could you please test against the production dump?


frcroth (approved):
works for me and should not break too much if it doesn't work

PR #393, opened 2023-10-11
do not update users from OIDC

### LINKED ISSUE #392: Do not override names for existing users
As the FSR, I want to be able to enter name changes into myHPI. Currently this is not possible as the values are overwritten with the values from OIDC on every login

jeriox:
closes #392


frcroth (commented):
Not completely happy approving this, since one of the big painpoints in 1327 was the constant stream of problems of people who dropped the student from their email (becoming phd students)

jeriox:
I discussed this with @dasGoogle and we thought that this PR shouldn't change anything about that, as the user with the changed mail adress wouldn't be found/authenticated in the first place

dasGoogle:
We will have to rewrite the whole authentication and user matching routine as soon as the new SSO is rolled out. There still is the discussion of whether to introduce an FSR SSO or directly attach to the new HPI SSO which will provide per-user UUIDs that should be static and can be used as an identifier instead of the email address.

dasGoogle:
Having talked to @lukasrad02 about this: The sub currently is firstname.lastname which would be static over an email address change. So there might be some value in still updating the email address on every sign in.

frcroth:
So update email address but do not update names? Would be fine with me. Also, add a comment with a link to this PR in the function.

frcroth (approved):

PR #406, opened 2023-10-17
Downgrade to Python 3.11 in Dockerfile

### LINKED ISSUE #404: Docker image build doesn't work: Failed to build uwsgi
https://github.com/fsr-de/myHPI/actions/runs/6539549001/job/17757877414

frcroth:
Python 3.12 just released (https://www.python.org/downloads/release/python-3120/) which broke our Dockerfile. I have downgraded to Python 3.11 for now, which is still supported. Of course in the long term, we will want to upgrade.

Closes #404


Paula-Kli:
I think it could be beneficial to also update the python version in the pre-commit hook as well - what do you think?

jeriox (approved):

PR #419, opened 2023-11-04
Use django bootstrap icon cache

frcroth:
Wow. For my local setup before this change the minimum page load time of the root page was abot 1200ms. With this change (after the first load) it is consistently between 200ms - 400ms. This is huge. :rocket:


jeriox:
@frcroth wow, amazing! I would be fine with approving this as a hotfix, but I think it would be nicer to download the icons once in the build step as we tried to avoid CDNs for bootstrap itself and i think it's weird that we then use a CDN for the icons.

frcroth:
Will you approve then? That is very low priority for me.

jeriox:
yes, but I would propose that you don't close the issue then so that we remember that there is something that can be done

jeriox (approved):

PR #421, opened 2023-11-09
Use tomselect

### LINKED ISSUE #336: Use select2 widget for author and moderator of minutes
Currently, you need to search in the simple select widget in a list of a few thousand people that aren't sorted alphabetically. We should use a select2 widget with a search function instead.

### LINKED ISSUE #411: Use tomselect instead of select2
If select2 doesn't work on newer wagtail versions, we should switch to an alternative.

See also: https://github.com/e-valuation/EvaP/pull/1758

frcroth:
Fixes #411 and fixes #336


jeriox (approved):
lgtm!
Could you use the display name of the user instead of the username? otherwise we need to change that again with the upcoming keycloak migration as the username will be the UUID from the OIDC sub claim in the future

myhpi/core/models.py
```diff
@@ -22,6 +23,36 @@
 from myhpi.core.widgets import AttachmentSelectWidget
 
 
+class UserProfile(models.Model):
```
jeriox, line 26: I personally don't really like the proxy models and generally prefer substituting the user model as a whole. However, this seems to be a bit tricky for existing projects (https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#changing-to-a-custom-user-model-mid-project).

jeriox (changes requested):
currently blocked as we cannot use first+last name as label

frcroth:
Also created an issue for that: https://github.com/OmenApps/django-tomselect/issues/7

jeriox:
we looked at the code of django-tomselect and it turns out that they pass the label field name to their own js code, where it is templated to access the data in some js object. so calling a function/property on the python model is definetly not possible and using a composite label also seems quite hard. I'd vote to abandon the tomselect plan

frcroth:
There has been a response: https://github.com/OmenApps/django-tomselect/issues/7#issuecomment-2546279475

PR #422, opened 2023-11-10
migrate to keycloak as IDP

jeriox:


frcroth:
Can you provide test setup instructions?

jeriox:
My testing looked as follows:
- log in with old OIDC provider with the configuration from main
- checkout this branch, enter client_id and secret from auth.myhpi.de in .env
- login with Keycloak and see how the sub of your user changes from firstname.lastname to your UUID

dasGoogle:
As we now have set up some basic group management in our Keycloak, I'd suggest we also integrate that into this PR. 

@frcroth In case you want to test the setup yourself, please contact me or @lukasrad02 directly so we can give you some testing credentials.

jeriox:
@dasGoogle yep that was the plan, can you provide me with some infos on how the groups look in the OIDC claim or do I have to figure that out on my own? :D (would probably be good to write that down somewhere for the future anyways)

jeriox:
After another discussion with @dasGoogle we decided against a separate identifier for the groups and will instead use the group name directly (e.g. "student"). We could change our templates to capitalize the group names for the UI

lukasrad02 (approved):

lukasrad02 (approved):

PR #425, opened 2023-11-21
Localize last edited time on information pages

### LINKED ISSUE #322: Last edited by uses UTC
Should use users' local time

image
image

dropforge:
Use the browser's locale to format the date and time of the last edit. If JavaScript is disabled, show "UTC" after the time.

Fixes #322


myhpi/core/templates/core/information_page.html
```diff
@@ -17,9 +18,22 @@ <h1 class="page-title">
...
+                <time id="last-published" datetime="{{ page.last_published_at|date:"c" }}" title="{{ TIMEZONE }}">
+                    <span id="last-published-date">
+                        {{ page.last_published_at|localtime|date:"SHORT_DATE_FORMAT" }}
+                    </span>
+                    {% translate "at" %}
+                    <span id="last-published-time">
+                        {{ page.last_published_at|localtime|time:"TIME_FORMAT" }}
+                    </span>
+                {% get_current_language as LANGUAGE_CODE %}
+                {% if LANGUAGE_CODE != "en" %}
```
jeriox, line 31: You don't need to think about the current_language in templates. Just use the `translate` template tag with the english value and django will take care of displaying the correctly translated output

myhpi/core/templates/core/information_page.html
```diff
@@ -17,9 +18,22 @@ <h1 class="page-title">
         <div class="side-panel-container">
             <aside class="side-panel border-accent">
                 <h1 class="toc-title">{% translate "Last edited" %}</h1>
-                {{ page.last_published_at|date:"d.m.Y, H:i" }}
+                {% get_current_timezone as TIMEZONE %}
+                <time id="last-published" datetime="{{ page.last_published_at|date:"c" }}" title="{{ TIMEZONE }}">
+                    <span id="last-published-date">
+                        {{ page.last_published_at|localtime|date:"SHORT_DATE_FORMAT" }}
```
jeriox, line 24: no need for localtime, tz_aware times (which those should be) will automatically be converted by django

myhpi/static/js/myHPI.js
```diff
@@ -54,6 +54,29 @@ const toggleElementVisibilityOnScroll = (minScrollPosition = 0) => {
     previousScrollPosition = currentScrollPosition
 }
 
+const localizeLastPublished = () => {
```
jeriox, line 57: Django supports activating a current timezone for the user natively, see https://docs.djangoproject.com/en/4.2/topics/i18n/timezones/#selecting-the-current-time-zone

I would prefer using that over a JS fix

jeriox (changes requested):

myhpi/core/templates/core/information_page.html
jeriox: You could use a blocktranslate tag and then use the value inside of that. In the german version you could then just add the "Uhr" behind the value

myhpi/static/js/myHPI.js
jeriox: thats not possible, as the browser does not communicate the local timezone to the server. However, django would then use the default timezone. For me that behaviour would be fine

jeriox (approved):

PR #432, opened 2023-12-20
Hotfix: Update username to be able to log in a second time

lukasrad02:
Due to the migration, the claims contain changed usernames and email addresses. However, only the new email address was preserved.
Therefore, the account was not found during a second log in.


dasGoogle (approved):
This should fix the issue based on the manual testing I just did.

dasGoogle:
As this is time-critical, we are merging without another review.

PR #437, opened 2023-12-20
Open external links in another tab and add icon showing that it's an external link

### LINKED ISSUE #212: Open external links in new tab by default
> Opening external pages in a new tab would be pretty easy to do with your external filter.

External links should open in a new tab by default. Adapt the `tag_external_links` in `core_extras.py` so it inserts the `target="_blank"` into the link tag.

_Originally posted by @SilvanVerhoeven in https://github.com/fsr-de/myHPI/issues/198#issuecomment-1290727749_

abc013:
Closes #212.

### DIFF
--- a/myhpi/core/templatetags/core_extras.py
+++ b/myhpi/core/templatetags/core_extras.py
@@ -38,10 +38,14 @@ def insert_footer(page):
 @register.filter(name="tag_external_links")
 def tag_external_links(content):
     """Takes the content of a website and inserts external link icons after every external link."""
-    external_links = re.finditer('<a[^>]*href="(?!' + settings.SITE_URL + ")[^>]*>[^<]*", content)
+    external_links = re.finditer(
+        '<a[^>]*href="(?!' + settings.SITE_URL + ")(?!\/)[^>]*>[^<]*", content
+    )
     for link in reversed(list(external_links)):
         content = (
-            content[: link.end()]
+            content[: link.start() + 2]
+            + " target='_blank'"
+            + content[link.start() + 2 : link.end()]
             + " {% bs_icon 'box-arrow-up-right' extra_classes='external-link-icon' %}"
             + content[link.end() :]
         )
--- a/myhpi/core/templates/core/information_page.html
+++ b/myhpi/core/templates/core/information_page.html
@@ -12,7 +12,7 @@
                 {{ page.title }}
             </span>
         </h1>
-        {{ parsed_md.0 }}
+        {{ parsed_md.0|tag_external_links }}
     </div>
     <div class="col-lg-3">
         <div class="side-panel-container">
--- a/myhpi/core/templates/core/minutes.html
+++ b/myhpi/core/templates/core/minutes.html
@@ -11,7 +11,7 @@
             <h1>{{  page.title }}</h1>
         </div>
         <div class="col-9 minutes-text">
-            {{ parsed_md.0 }}
+            {{ parsed_md.0|tag_external_links }}
         </div>
         <div class="col-3 minutes-meta">
             <aside class="side-panel border-accent">


myhpi/core/markdown/extensions.py
```diff
@@ -160,6 +160,21 @@ def url(self, id):
...
 
+class ExternalLinkPattern(LinkInlineProcessor):
+    def handleMatch(self, m, data=None):
+        el = util.etree.Element("a")
+        try:
+            el.set("href", m.group("url"))
+            el.set("target", "_blank")
+            el.set("class", "external-link-icon")
+            el.text = util.AtomicString(m.group("title"))
+        except ObjectDoesNotExist:
```
jeriox, line 171: this cannot happen as you don't do a database lookup

myhpi/core/templatetags/core_extras.py
```diff
@@ -35,20 +35,6 @@ def insert_footer(page):
     return {"footer_columns": [footer.column_1, footer.column_2, footer.column_3], "page": page}
 
 
-@register.filter(name="tag_external_links")
-def tag_external_links(content):
-    """Takes the content of a website and inserts external link icons after every external link."""
-    external_links = re.finditer('<a[^>]*href="(?!' + settings.SITE_URL + ")[^>]*>[^<]*", content)
```
jeriox, line 41: this also checked whether the link is part of our page by comparing to `settings.SITE_URL`. You could also check that, as a manual link (e.g. if I entered "myhpi.de/fachschaftsrat" into a page) would also be labeled as external if I understand your code correctly

myhpi/core/markdown/extensions.py
```diff
@@ -160,6 +160,21 @@ def url(self, id):
     def default_pattern():
         return r"\[(?P<title>[^\[]+)\]\(page:(?P<id>\d+)\)"
 
+class ExternalLinkPattern(LinkInlineProcessor):
+    def handleMatch(self, m, data=None):
+        el = util.etree.Element("a")
+        try:
+            el.set("href", m.group("url"))
```
jeriox, line 167: is it necessary to re-set the href here? maybe you should check the priority of markdowns built-in link parser and whether you want to replace or supplement it

myhpi/templates/footer.html
```diff
@@ -6,7 +6,7 @@
     {% for column in footer_columns %}
     <div class="footer-category">
       {% with column|markdown as parsed_markdown %}
-      {{ parsed_markdown.0|tag_external_links }}
+      {{ parsed_markdown.0 }}
```
jeriox, line 9: doesnt't that mean that you could just send the body of a page through the same old filter and the issue would be resolved?

abc013: True that. I'll take the old filter then (soon™)

abc013: Very fair argument, although i wonder to what extend i should apply this filter (alias, where should i put it apart from the content block?)

Right now, that very filter is also marking local links in the sidebar
image
But i guess that's because the filter checks for `settings.SITE_URL`, so it shouldn't be a problem in production, right? Still, this feels quite sketchy imo.

jeriox: I don't think that you should apply it to ToC, as this only renders headings and headings should never be hyperlinks (apart from the ToC where you navigate to the anchor). But if you only apply it to the main content it should be fine I think?

abc013: Sadly, it also shows internal links in the main content as external. Probably, this works in production but probably won't in local setups...

jeriox: yeah I don't think that this can be fixed apart from setting the SITE_URL

abc013:
Updated.

jeriox (approved):

PR #438, opened 2023-12-20
Instant run-off elections

### LINKED ISSUE #81: Alert Close Button is ugly and does not work
As can be seen in the screenshot, the button does not fit with its background. Additionally, the button does not actually remove the alert. 
image

### LINKED ISSUE #220: Allow Ranking of poll options

BenBals:
closes #220 closes #81


frcroth:
Can you merge main into this? Makes testing easier

myhpi/polls/models.py
```diff
@@ -18,24 +22,50 @@ class PollList(BasePage):
...
+        "PollList",
+    ]
+    subpage_types = []
+    is_creatable = False
+
+    def in_voting_period(self):
+        return self.start_date <= datetime.date.today() <= self.end_date
+
+    def can_vote(self, user):
+        return self.in_voting_period() and user not in self.already_voted.all()
```
frcroth, line 55: I would like the poll to also have a list of groups I can limit it to. Otherwise profs etc. can also vote (I think this should be different from visibility, because visibility may not limit POSTing responses)

myhpi/polls/templates/polls/ranked_choice_poll.html
```diff
@@ -0,0 +1,51 @@
+{% extends "polls/base_poll.html" %}
+{% load core_extras %}
+{% load core_extras %}
+{% load i18n %}
+
+{% block ballot %}
```
frcroth, line 6: This page looks ok for the first version I think.

frcroth (commented):
- Poll list could look better:
image
- For me the poll list is not shown in the menu?
What should I review here if the calculation is not in yet?

jeriox:
>     * For me the poll list is not shown in the menu?

did you tick the "show in menus" checkbox? otherwise probably menu caching again

>       What should I review here if the calculation is not in yet?

data model, vote casting, esp. if there are some cases that I missed, like the one with restricting POSTing that you already mentioned

myhpi/polls/models.py
```diff
@@ -18,75 +22,96 @@ class PollList(BasePage):
...
-                    self.participants.add(request.user)
-                    messages.success(request, "Your vote has been counted.")
-                return redirect(self.relative_url(self.get_site()))
+    is_creatable = True
 
-        return super().serve(request, *args, **kwargs)
+    def cast_vote(self, request, *args, **kwargs):
+        choices = request.POST.getlist("choice")
+        if len(choices) == 0:
+            messages.error(request, "You must select at least one choice.")
```
frcroth, line 95: Any way to create translations for these?

myhpi/polls/models.py
```diff
@@ -98,26 +123,174 @@ def num_votes(self):
...
+    is_creatable = True
+
+    def cast_vote(self, request, *args, **kwargs):
+        form = self.get_ballot_form(request.POST)
+        if not form.is_valid():
+            messages.error(
+                request, mark_safe(_("Invalid ballot.\n{errors}").format(errors=form.errors))
+            )
+        else:
+            try:
```
frcroth, line 172: Can I theoretically send multiple valid form responses in the same moment and create a condition where I would be able to vote multiple times (since already_voted is only set at the end)?

myhpi/polls/models.py
```diff
@@ -98,26 +123,174 @@ def num_votes(self):
...
+
+    def as_sensible_datastructure(self):
+        result = []
+        for entry in self.rankedchoiceballotentry_set.all():
+            result.append((entry.rank, entry))
+        heapq.heapify(result)
+        return result
+
+    def __str__(self):
+        return "ballot"  # , ".join(map(lambda x: str(x), self.entries.all()))
```
frcroth, line 280: Remove comment or method if not necessary

myhpi/tests/polls/test_ranked_choice_algorithm.py
```diff
@@ -0,0 +1,107 @@
...
+            path="0001000200010005",
+            depth=4,
+            is_public=True,
+        )
+        self.information_menu.add_child(instance=self.poll_list)
+
+        self.poll = RankedChoicePoll(
+            title="How are you?",
+            slug="how-are-you",
+            description="This is a poll to check how you are.",
```
frcroth, line 45: Adjust text for the options to make more sense

myhpi/tests/polls/test_ranked_choice_algorithm.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+    def test_no_ballots(self):
+        self.assertEqual(
+            self.poll.calculate_ranking(),
+            [(1, "Alice", 0), (1, "Bob", 0), (1, "Charlie", 0), (1, "Dora", 0)],
+        )
+
+    def test_fast(self):
+        # TODO: Please assert this is reasonably fast. Do your Django magic here.
```
frcroth, line 97: Who is this addressed to?

frcroth (commented):
Nice, great work. I have some minor things:

- When submitting invalid responses, the message has weird text
image
- Could add some spacing between the vote button and the footer
image
- You could add poll list and polls to the setup test data script (or create a followup issue for that)

myhpi/polls/models.py
BenBals: Hmm, I don't know much about Django ORM, but could we maybe move the uniqueness of votes constraint to the database, perform the insert of the user before the insert of the ballot and error out on a double insertion of the first? 

This way, we wouldn't have to worry about read-write locking the voter database.

myhpi/tests/polls/test_ranked_choice_algorithm.py
BenBals: Fixed by eebe3ce9c8b24b48c6e0f6ccd5e9cf13e7454835

myhpi/tests/polls/test_ranked_choice_algorithm.py
BenBals: I already asked both DuckDuckGo and @jeriox, but got no better answer then just timing it, so I just implemented that in 9a546d5024c470c177e6fb8dca354a3fe4dff3d3

Though, if you have a better idea, I'd also try that.

BenBals:
What's left to do:

`as_sensible_datastructure` performs an individual database request for each and every ballot. We expect vote numbers to be (at most) in the low thousands. Still, we should prefetch this data in `calculate_votes`. 

Relevant code:
```python
    def as_sensible_datastructure(self):
        result = []
        for entry in self.rankedchoiceballotentry_set.all():
            result.append((entry.rank, entry))
        heapq.heapify(result)
        return result
```

and 

```python
    def calculate_ranking(self):
        ballots = list(map(lambda x: x.as_sensible_datastructure(), list(self.ballots.all())))
        options = self.options.all()
        [snip]
```

jeriox:
>     * Could add some spacing between the vote button and the footer
>       image

@SilvanVerhoeven

myhpi/tests/polls/test_ranked_choice_algorithm.py
```diff
@@ -0,0 +1,123 @@
...
+        self.poll_list.add_child(instance=self.poll)
+
+    def cast_ballots(self, template_ballots):
+        for template_ballot in template_ballots:
+            ballot = RankedChoiceBallot.objects.create(poll=self.poll)
+            ballot.save()
+
+            for idx, template_entry in enumerate(template_ballot):
+                RankedChoiceBallotEntry.objects.create(
+                    ballot=ballot, option=self.options[template_entry], rank=idx
```
frcroth, line 64: Maybe using the POST route would be nice here? That would actually test some of the behavior there. Maybe a test would then be possible where two requests are sent at the same time, though I am actually not sure if that is possible to be implemented in a non-flaky way.

myhpi/polls/models.py
```diff
@@ -98,26 +123,188 @@ def num_votes(self):
...
+                mark_safe(
+                    _("Invalid ballot.\n{errors}").format(
+                        errors=", ".join([str(err) for err in form.errors.values()])
+                    )
+                ),
+            )
+        else:
+            try:
+                self.already_voted.select_for_update(nowait=True).get(user=request.user)
+                with transaction.atomic():
```
frcroth, line 179: Rather lock the whole poll with wait

myhpi/polls/models.py
```diff
@@ -98,26 +123,188 @@ def num_votes(self):
...
+                assigned_rank = rank
+            results.append((assigned_rank, names[candidate[1]], candidate[0]))
+            previous_votes = candidate[0]
+
+        return sorted(results)
+
+
+class RankedChoiceOption(Orderable):
+    name = models.CharField(max_length=254)
+    description = CustomMarkdownField()
```
frcroth, line 277: Would be fine with making this optional

frcroth (commented):
Translations yet to be added?

myhpi/polls/models.py
jeriox: we decided to keep it required

frcroth (approved):
:sweat_smile:

PR #452, opened 2023-12-31
Add dynamic versioning and version in footer

### LINKED ISSUE #400: Print last commit hash on page
For debugging, it would be nice to see what build is used in the running deployment, which may be printed somewhere on the page.

frcroth:
Fixes #400

For me this does not yet work locally (The version is always 0.1.0) even though I basically used the ephios code. Do you have an idea here @jeriox ? `poetry build` produces a dynamic version number


myhpi/core/migrations/0002_rootpage.py
```diff
@@ -50,7 +50,7 @@ def create_homepage(apps, schema_editor):
     Footer.objects.get_or_create(
         column_1="# Fachschaft\r\n\r\n- [Twitter](https://twitter.com/fachschaftsrat)\r\n- [Discord](https://discord.com)\r\n- [Telegram](https://telegram.org)",
         column_2="# Rechtliches\r\n\r\n- [Impressum]()\r\n- [Datenschutzerklärung]()",
-        column_3="# Entwicklung\r\n\r\n- [GitHub](https://github.com/fsr-de/myHPI/)",
+        column_3="# Entwicklung\r\n\r\n- [GitHub](https://github.com/fsr-de/myHPI/)\r\n[MYHPI-VERSION]",
```
jeriox, line 53: maybe we should re-visit how the footer is built. I already removed the last column for the language as this is handled programatically, now we have this variable here where the preprocessor seems like a very complicated way to achieve this. But not part of this PR

jeriox (approved):

jeriox:
> For me this does not yet work locally (The version is always 0.1.0) even though I basically used the ephios code. Do you have an idea here @jeriox ? `poetry build` produces a dynamic version number

sounds like you have an "old" version installed in your venv that takes precedence for the version. Did you run poetry install afterwards?

frcroth:
> sounds like you have an "old" version installed in your venv that takes precedence for the version. Did you run poetry install afterwards?

Nice, that's it!

myhpi/core/migrations/0002_rootpage.py
frcroth: #455

PR #456, opened 2024-01-03
Publish docker images with labels

### LINKED ISSUE #426: Tag docker images with versions
To create docker images with versioning so that @lukasrad02 doesn't have to use :latest tag in a docker file, which can make it difficult to replicate a deployment.

May be related to #400

frcroth:
Possibly Fix #426 

Based on https://stackoverflow.com/questions/66017161/add-a-tag-to-a-docker-image-if-theres-a-git-tag-using-github-action


jeriox:
What is the metadata-action used for? (comparing against https://github.com/ephios-dev/ephios/blob/main/.github/workflows/docker.yml). And you probably need the permission stuff to write packages as well

jeriox:
We create releases in the GitHub webinterface, and this automatically triggers the workflow and creates the correctly tagged packages (https://github.com/ephios-dev/ephios/pkgs/container/ephios)

jeriox (approved):

PR #492, opened 2024-02-13
Show date for search results

### LINKED ISSUE #399: Display Minutes Date in Search Mode
In the search mode of the minutes list it would be great to see also the date the minute was created and not only the title as the title might be repeatedly used

dropforge:
Shows the localized date on the right for search results. Closes #399 

results-date


myhpi/search/templates/search/search_result.html
```diff
@@ -3,7 +3,12 @@
 
 <a href="{% pageurl result %}">
     <div class="search-result">
-        <h5>{{ result.title }}</h5>
+        <hgroup>
+            <h5>{{ result.title }}</h5>
+            <time datetime="{{ result.first_published_at|date:"c" }}">
+                {{ result.first_published_at|date:"SHORT_DATE_FORMAT" }}
```
jeriox, line 9: this is actually a bit misleading (our bad, the issue states it like you implemented it). We usually only publish our minutes about 5 or 6 days after the actual meeting, after everyone that participated had a chance to review it. Therfore, minutes have an additional field carrying the date of the meeting. My preferred choice would be to only display this date for minutes and nothing for other types of pages, as the date isn't really helpful there IMO

myhpi/static/scss/myHPI.scss
```diff
@@ -181,6 +181,15 @@ h1.side-panel-title::after {
     margin: -0.5rem -0.8rem;
 
     .search-result {
+        hgroup {
+          display: flex;
+          justify-content: space-between;
```
SilvanVerhoeven, line 186: So far we've tried to keep our scss to a minimum and instead use bootstrap's utility classes. You could use `d-flex justify-content-between`

SilvanVerhoeven (changes requested):

myhpi/search/templates/search/search_result.html
```diff
@@ -3,7 +3,12 @@
 
 <a href="{% pageurl result %}">
     <div class="search-result">
-        <h5>{{ result.title }}</h5>
+        <hgroup class="d-flex justify-content-between">
+            <h5 class="text-break">{{ result.title }}</h5>
+            <time datetime="{{ result.specific.date|date:"c" }}">
+                {{ result.specific.date|date:"SHORT_DATE_FORMAT" }}
```
frcroth, line 9: Since other pages don't have date this is now only for minutes? Intentional?

SilvanVerhoeven (approved):
Aside from @frcroth's question lgtm

myhpi/search/templates/search/search_result.html
lukasrad02: Yes, see <https://github.com/fsr-de/myHPI/pull/492#discussion_r1490861347>

jeriox (approved):

PR #496, opened 2024-02-13
Improve TOC heading spacing

### LINKED ISSUE #495: Better multiline headings in TOC
In a page's table of contents, multi-line headings are difficult to discern. Instead, they should be grouped together.

image

dropforge:
Remove spacing between lines of the same multi-line heading and add spacing to the next item in the table of contents. Closes #495

image


myhpi/static/js/myHPI.js
```diff
@@ -70,6 +70,7 @@ const localizeLastPublished = () => {
         let abbr = document.createElement("abbr");
         abbr.setAttribute("title", lastPublishedLocalized)
         lastPublished.removeAttribute("title")
+        lastPublished.innerHTML = lastPublished.textContent.trim()
```
jeriox, line 73: why is this change necessary?

jeriox (approved):

myhpi/static/js/myHPI.js
frcroth: Does this also work with innerText to get rid of this security warning?

PR #504, opened 2024-02-13
Resolution button for minute editor

### LINKED ISSUE #335: Add button for resolutions in markdown editor
Noting down Beschlüsse is not obvious

dropforge:
For expenses there a vote is held and it must be categorized into a budget. Moves the vote results to the end of the line. Closes #335 

image


myhpi/static/js/admin/easymde_custom.js
```diff
@@ -157,3 +163,16 @@ function addQuorum(editor){
     output = "\n|quorum|(/)";
     cm.replaceSelection(output);
 }
+
+function addResolution(editor){
+    const cm = editor.codemirror;
+    const position = cm.getCursor();
+
+    const output = " **[//]** (HHT: ).";
```
jeriox, line 171: You need to ouput | (pipe), otherwise the resolution will not be highlighted. I would also prefer if this button also outputted the actual resolution text ("Wir beschließen bis zu x€ für y auszugeben"

jeriox: when rendering the markdown to html, blocks with something like [1|2|3] will be highlighted, that was what I orignally meant. 

the other thing is how the content of the commands is actually stored. All the other buttons just output smth like `|start|(15:24)` to reduce duplication of the actual text snippets which people could mistype. A markdown processor (have a look at extensions.py) then takes this snippet and replaces it with the actual Text ("Sitzungsbeginn 15:24") in the correct locale. You could implement the resolution in the same way (e.g. `|resolution|(ding)(betrag)[1|2|3]`)

jeriox: we will do another release as soon as your PRs are merged :)

myhpi/core/markdown/extensions.py
```diff
@@ -97,6 +97,24 @@ def quorumify(self, match):
         )
 
 
+class ResolutionPreprocessor(MinutesBasePreprocessor):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.patterns = [
+            (r"\|resolution\|\(\d+([.,]\d{2})?\)\(\S+\)\(\S\)\[(\d+)\|(\d+)\|(\d+)\]", self.resolutify),
```
jeriox, line 104: in my local testing, this regex did not match the string that the button inserted into the editor (and me filling in values)

jeriox (changes requested):

frcroth (commented):
Can you add tests :point_right: :point_left:

frcroth (approved):

PR #505, opened 2024-02-13
Create a New Admin Widget for Text Arrays

### LINKED ISSUE #332: Use better widget for minutes guests in wagtail admin
Currently, the guest field on minutes is using a simple textarea as a widget which displays a JSON array of strings representing the guests. There should be an easy to use widget, like in 1327 with a "add more guests" button. IMO we already implemented this here, maybe it broke somewhere along the way

DrEGZo:
This will fix #332.

Instead of a JSON string, I introduced a new Text Array Widget that can be used for lists of single-line strings such as the guest list.

You can add and delete text fields by clicking on the appropriate buttons. If you press `Enter` while typing in one of the text fields, a new text field at the end is created (and focussed). You can use `Tab` and `Shift+Tab` to switch between text fields. 

Empty text fields are stripped prior to saving.


frcroth (approved):
Awesome! :shipit: 

DrEgzo is valid JSON!

myhpi/core/templates/core/text_array_widget.html
```diff
@@ -0,0 +1,18 @@
+<link type="text/css" href="/static/css/text_array_widget.css" rel="stylesheet">
+<script src="/static/js/admin/text_array_widget.js"></script>
```
jeriox, line 2: I would prefer having these included as django media (https://docs.djangoproject.com/en/5.0/topics/forms/media/#assets-as-a-static-definition)

jeriox (approved):

myhpi/core/templates/core/text_array_widget.html
DrEGZo: Check out [https://github.com/fsr-de/myHPI/pull/505/commits/4015ded51604ef2b0a6bcf61829946ee8a358862](https://github.com/fsr-de/myHPI/pull/505/commits/4015ded51604ef2b0a6bcf61829946ee8a358862) and feel free to resolve this conversation and merge :)

PR #507, opened 2024-02-13
Add Newsfeed

### LINKED ISSUE #479: Newsfeed
The current start page is quite boring. Let's make it more interesting! How about a feed of information that is relevant to the user?
- Announcements by the FSR, StuKo...
- New myhpi release announcements
- Announcements by other groups the user is a member of (e.g. Klubs)
- Maybe new minutes released (Also mentioned in #17)

The news articles could be a new type of page. I am imagining a route for creating them that could also be automatically filled. So for example if the FSR sends a mail to student, a new post is automatically created; if a myhpi release is published, this post is automatically created.

frcroth:
Close #479 

image

# TODO
- [x] When logged in as student and going to the "/" page, redirect to the feed
- [x] Automatically write to feed on release. See https://github.com/frcroth/myHPI/actions/runs/7907525385/job/21584753770 for this. Obviously I couldn't test this properly (also the first time this will fail because the feed will only be deployed after the release is published).
- [x] Create tests
- [x] Add feed to setup test data
- [x] Pagination
- [x] Render html properly in snippet
- [x] Properly break snippet so that no html tag is unclosed
- [x] Translations
- [x] Ensure everything looks fine for mobile
- [x] Position pagination buttons better

## Optional todos (can also be followups)

- [ ] Rss feed
- [x] Icons for post accounts (Fsr logo for fsr posts)
- [ ] Non-optional but not really part of this PR: The thing that posts the FSR mails
- [ ] Reactions on posts?

# Questions
- how exactly should the redirect behavior work. Always redirect when clicking on "myhpi"? Redirect only when logging in and no next is set?
- 
@jeriox we can do the election release first and merge this afterward if there were no problems


myhpi/feed/middleware.py
```diff
@@ -0,0 +1,26 @@
+from django.http import HttpResponseRedirect
+
+from myhpi import settings
+from myhpi.feed.models import NewsFeed
+
+
+class FeedRedirectMiddleware:
```
jeriox, line 7: you could also implement this inside the serve() of RootPage, that would only do the check when it is actually necessary and not with every request

myhpi/feed/models.py
```diff
@@ -0,0 +1,74 @@
...
+        # if user:
+        #    profile = UserProfile.get_for_user(user)
+        #    return profile.avatar.url
+        else:
+            return "https://www.gravatar.com/avatar/b1a83a1baa8a1d45c905a59217a7d30a?s=140&d=mm"
+
+
+# The PostAccount is used to create posts via API
+@register_snippet
+class PostAccount(models.Model):
```
jeriox, line 67: I'm not sure if i like the name, maybe more like NewsfeedAccount?

myhpi/feed/models.py
```diff
@@ -0,0 +1,74 @@
...
+    post_key = models.CharField(max_length=255)
+    name = models.CharField(max_length=255)
+
+    icon = models.ImageField(upload_to=upload_avatar_to, null=True, blank=True)
```
jeriox, line 71: is this the correct reference to also use all of the wagtail stuff for images?

myhpi/feed/views.py
```diff
@@ -0,0 +1,38 @@
...
+import json
+
+from django.core.exceptions import MultipleObjectsReturned, SuspiciousOperation
+from django.http import HttpResponse
+from django.views import View
+
+from myhpi.feed.models import NewsFeed, Post, PostAccount
+
+
+class PostFeedEntryView(View):
```
jeriox, line 11: is there a reason why you are not using the [wagtail API](https://docs.wagtail.org/en/stable/advanced_topics/api/v2/configuration.html) for this?

dasGoogle:
As an idea on how to handle this on the main page: Provide the first 3-4 news items (I think we could go for a two-column layout there and maybe even make the design a bit more compact) and add a link to "Further News" or similar.

myhpi/feed/middleware.py
frcroth: But that would increase coupling. Currently disabling/removing the app is very easy, it is never mentioned / used in core.

myhpi/feed/models.py
frcroth: This is the same code that is used for uploading user avatars in Wagtail. Does not have to deal with collections etc., which makes sense because the post accounts should only be edited by admins anyway.

myhpi/feed/views.py
frcroth: Did not find them when googling (searched for creating a django API). Will look into it.

frcroth: Wagtail API is read only, so does not fit the use case.

lukasrad02:
Regarding the *Automatically write to feed on release* feature: Do you think it would make sense to integrate this into the myHPI source code instead of the GitHub pipeline? I would think about some logic that checks the current version on startup, compares it with the version from the last startup and fetches the release description from GitHub if the version has changed.

I would prefer this over creating the post from the pipeline, as there are usually a couple of hours or days between the release on GitHub and deploying it in production. Having some large post advertising new features on the homepage although these features are not even available might confuse users.

lukasrad02:
You've got a good point. I was thinking about comparing the version numbers, but this would of course require us to store them in addition to the posts, like the separate structure you've mentioned.

I don't have a strong opinion on whether to introduce such a structure or not, but am slightly in favor of doing this. Otherwise, we would even have to differentiate between release posts and other posts during diffing, so there will be a lot of edge cases to consider.

dasGoogle:
As another Idea: Why don't we pull the releases from GitHub in the action and place it in a file in the docker container? 
We could then simply parse that file on startup and mix it into the existing newsfeed without persisting anything in the DB.

jeriox:
I think all of your variants are quite complex. I would prefer just using the tag of the current version (which we have easy acccess to from the python code) and creating a post for that by pulling the corresponding release notes if it does not exist yet.

lukasrad02:
From my perspective, both the suggestions from Aaron and Julian would be fine

frcroth:
Now implemented @jeriox 's version. the only problem I see here is that some releases may not be published. I.e.
- New version 2.0 is published with many cool features.
- A bug is found immediately after publishing (or while deploying)
- Version 2.0.1 is published with one fix.
- It deploys successfully, and only the 2.0.1 post is published.

To fix this, an admin needs to manually publish the post with `python manage.py publish_release_post v2.0.0`

Coming back to how the user is presented the feed. What I like about the current version is that no changes are done to core. I am not sure if that would be possible when including the posts on the root page. If that is not relevant, I think it should be possible.

myhpi/feed/middleware.py
lukasrad02: In my opinion, it would be okay to introduce some change to core, as myHPI (this application) is developed specifically for our myHPI website and not meant as a generic CMS. Hence, we don't need to be able to deactivate/remove the newsfeed app per installation and that small amount of coupling would be acceptable.

However, I'm not that familiar with Django patterns, so I can't really argue on whether there is even a need to replace this middleware or not.

lukasrad02:
> To fix this, an admin needs to manually publish the post with `python manage.py publish_release_post v2.0.0`

Nice solution!
 
> Coming back to how the user is presented the feed. What I like about the current version is that no changes are done to core. I am not sure if that would be possible when including the posts on the root page. If that is not relevant, I think it should be possible.

Added some comment on this to the discussion above.

myhpi/feed/middleware.py
jeriox: @lukasrad02 Felix and I had a conversation about the conflict of cool new features vs. size of codebase/maintainability. I made the argument that as long as fancy features are loosely coupled, future maintainer could easily remove them without the need to understand/update/fix them and therefore delay or block the maintenance of the core app.

The middleware is probably fine, as it doesn't do much. I just thought about it because we always struggle with pageload times and this is code that is not necessary on all pages but the root page.

Another way that we could pursue this is by adding some kind of "plugin" mechanism. Again citing ephios: we are sending out custom django signals to "plugins" (other django apps within the same repo or installed separately) and let them inject stuff at various places. e.g. there is a signal that says "hey plugin, give me some html to append to the homepage if you want to". This way, the additional code is only run when it is necessary.

lukasrad02: I'm fine with you two deciding on the best approach to do this. Especially when it comes to performance or plugins in Django, I would not be able to give any qualified arguments.

PR #515, opened 2024-02-16
Disable menu cache when in debug mode

### LINKED ISSUE #481: Disable menu caching in debug mode
When developing locally, the menu caching can lead to a lot of confusion. We should disable the caching if DEBUG=true

frcroth:
Closes #481

### DIFF
--- a/myhpi/core/context.py
+++ b/myhpi/core/context.py
@@ -40,7 +40,7 @@ def base_context(request):
         "root_page": root_page,
         "pages_by_parent": page_lookup,
         "minutes_creation_links": minutes_creation_links,
-        "enable_menu_cache": not settings.DEBUG,
+        "template_cache_duration": 1 if settings.DEBUG else 500,
     }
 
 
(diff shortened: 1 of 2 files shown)


myhpi/core/context.py
```diff
@@ -39,6 +40,7 @@ def base_context(request):
         "root_page": root_page,
         "pages_by_parent": page_lookup,
         "minutes_creation_links": minutes_creation_links,
+        "enable_menu_cache": not settings.DEBUG,
```
jeriox, line 43: you could also set the template_cache_duration here and use that in the template to avoid duplication and use a value of 0 or 1 seconds in debug mode

frcroth: you're right, did it now. Thinking about that: It could also just be a setting/.env key where the .env.example contains 1 and the production .env has a different value (e.g. 500)

jeriox (approved):

myhpi/core/context.py
frcroth: Does the thumbs up mean you want it as a settings key in this pr or not?

jeriox: If you want to include it in this PR go ahead, but I don't think it is necessary as we currently don't have any other usecase apart from the one in this PR

PR #526, opened 2024-02-23
use django management commands for tools

### LINKED ISSUE #445: Use django management commands for tools
See https://github.com/fsr-de/myHPI/pull/444#pullrequestreview-1795346225

Use django management commands for tools: https://docs.djangoproject.com/en/5.0/howto/custom-management-commands/

jeriox:
closes #445


Dockerfile
```diff
@@ -13,6 +13,6 @@ RUN pip install "poetry==$POETRY_VERSION" "uwsgi==$UWSGI_PIP_VERSION" "psycopg2=
 RUN poetry self add "poetry-dynamic-versioning[plugin]"
 ADD . /app
 RUN poetry install --no-dev
-RUN python tools/install_bootstrap.py -u
+RUN python tools/install_bootstrap.py -u # FIXME
```
jeriox, line 16: @frcroth django management command do a full setup before running (to make interacting with the database etc. easier). When building the container, we cannot do that because the environment to run django in is not there yet. This means we have to either revert my changes and keep the bootstrap installation as a standalone script or handle the bootstrap installation when deploying. What are your thoughts on this?

frcroth: I would prefer the bootstrap stuff to remain in tools in that case. I think it still makes sense there, because I would expect django management tools to interact with django and this tool doesn't really do that.

.github/workflows/tests.yml
```diff
@@ -73,22 +73,17 @@ jobs:
...
         sudo apt-get install gettext
         source env/bin/activate
         poetry run python -m pip install setuptools -U
         poetry install -E mysql -E pgsql
-        python tools/install_bootstrap.py
-    - name: Prepare files for test run
-      continue-on-error: true
-      run: |
         cp .env.example .env
-        source env/bin/activate
```
frcroth: If it works without sourcing the venv here, does it work in the other steps as well?

myhpi/core/management/commands/create_test_data.py
```diff
@@ -354,14 +342,18 @@ def create_some_pages(users, groups, documents):
...
-    create_abbreviation_explanations()
-    create_some_pages(users, groups, documents)
-
+class Command(BaseCommand):
+    help = "Creates test data (user, pages, etc.) for myHPI."
 
-if __name__ == "__main__":
-    main()
+    def handle(self, *args, **options):
+        # Superuser is created manually via createsuperuser command
```
frcroth, line 349: Did you test running it twice? Should it handle that in some way?

Dockerfile
jeriox: @frcroth fixed

frcroth (approved):

.github/workflows/tests.yml
jeriox: outdated

myhpi/core/management/commands/create_test_data.py
frcroth: Ok, feel free to merge

Issue #345, opened 2023-04-22
Feature Request: Make it easier to search for protocols


LinqLover:
Not sure whether this is a technical or organizational issue, but currently it is rather hard to search the protocols for specific terms because the search results seem to mix up younger and older results: https://myhpi.de/de/search/?query=professur

An option to search in specific folders or to order results by date descending/ascending would be great, but it would already be a help if each protocol item in the search restuls was displayed with a date.

frcroth:
Related to #399

jeriox:
fixed in #577

PR #557, opened 2024-06-10
Improve search results page

dasGoogle:
This is a PR to track some improvements to the search page. This includes proper sorting, a better content preview and maybe additional filters.

### DIFF
--- a/myhpi/search/templatetags/search.py
+++ b/myhpi/search/templatetags/search.py
@@ -2,6 +2,7 @@ import re
 from django import template
 from myhpi.core.markdown.utils import render_markdown
 from django.utils.safestring import mark_safe
+from django.utils.html import escape
 
 register = template.Library()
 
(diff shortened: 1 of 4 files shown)


dasGoogle:
Ping @jeriox @lukasrad02 for your suggestions

jeriox:
I think you dropped the date (https://github.com/fsr-de/myHPI/pull/492), or is your example not a protocol? I think we should definetly include that.

Could you also provide a screenshot with multiple results? With the current layout I suspect that it could be hard to visually distinguish between different results as they are not as uniform as before

dasGoogle:
image

Indeed, the screenshot was from a page with minutes content, the date remains in there for minutes

jeriox:
I like it

jeriox:
another suggestion from #345: restrict search to specific areas of the page (pages/minutes/even specific minutes lists or respective groups?) -> can also be a separate issue if you don't want to implement it in this PR

jeriox (approved):

myhpi/search/templatetags/search.py
```diff
@@ -0,0 +1,57 @@
...
+from django.utils.safestring import mark_safe
+
+from myhpi.core.markdown.utils import render_markdown
+
+register = template.Library()
+
+
+@register.filter(name="highlight_query_markdown")
+# select 3 lines around first match and highlight query match
+def highlight_query(content, search_query, surrounding_lines=1):
```
frcroth, line 14: This function seems quite complex, maybe it could be tested?

PR #577, opened 2024-07-13
Set language cookie when selecting language

### LINKED ISSUE #559: Store user language preference
I use an English system language but prefer myHPI in German. I would like to not have to switch to German everytime I visit the page.

E.g. the language preference could be stored in my browser.

frcroth:
Closes #559 

Using the set language view from Django: https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#the-set-language-redirect-view, integrated into the existing links.

Using a form with hidden fields to achieve that.

### DIFF
--- a/myhpi/templates/footer.html
+++ b/myhpi/templates/footer.html
@@ -13,14 +13,12 @@
     <div class="footer-category">
       <h2>{% translate "Language" %}</h2>
       {% for translation in page.get_translations %}
-      <form id="language-form-{{ translation.locale }}" action="{% url 'set_language' %}" method="post" style="display: none;">
+      <form id="language-form-{{ translation.locale }}" action="{% url 'set_language' %}" method="post">
         {% csrf_token %}
-        <input name="next" type="hidden" value="{{ redirect_to }}">
+        <input name="next" type="hidden" value="{{ translation.url }}">
         <input type="hidden" name="language" value="{{ translation.locale.language_code }}">
+        <button class="btn btn-link" type="submit">{{ translation.locale }}</button>
       </form>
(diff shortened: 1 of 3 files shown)


myhpi/templates/footer.html
```diff
@@ -12,9 +12,16 @@
     {% endfor %}
     <div class="footer-category">
       <h2>{% translate "Language" %}</h2>
-        {% for translation in page.get_translations %}
-            <a href="{{ translation.url }}">{{ translation.locale }}</a>
-        {% endfor %}
+      {% for translation in page.get_translations %}
+      <form id="language-form-{{ translation.locale }}" action="{% url 'set_language' %}" method="post" style="display: none;">
+        {% csrf_token %}
+        <input name="next" type="hidden" value="{{ redirect_to }}">
```
jeriox, line 18: where is redirect_to coming from? shouldn't that be translation.url?

myhpi/templates/footer.html
```diff
@@ -12,9 +12,16 @@
...
+        <input type="hidden" name="language" value="{{ translation.locale.language_code }}">
+      </form>
+      <a href="#" onclick="document.getElementById('language-form-{{ translation.locale }}').submit(); return false;">
```
jeriox, line 21: I thin I'd prefer using a button with the `btn-link` class to style it as a link so that it works without js

myhpi/templates/footer.html
frcroth: It was always empty, so it seemed to work without setting next, but I applied your suggestion now.

jeriox (approved):

PR #592, opened 2024-07-30
Responsive design for polls

### LINKED ISSUE #588: Responsive layout for poll pages
Introduce the responsive layout from #525 for poll pages

SilvanVerhoeven:
Closes #588 

grafik

### DIFF
--- a/myhpi/core/templates/core/minutes_list.html
+++ b/myhpi/core/templates/core/minutes_list.html
@@ -3,7 +3,11 @@
 {% load i18n %}
 
 {% block content %}
-<h1>{{ page.title }} {{ selected_year }}</h1>
+<h1 class="page-title">
+    <span class="underline">
+        {{ page.title }} {{ selected_year }}
+    </span>
+</h1>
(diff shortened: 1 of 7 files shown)


SilvanVerhoeven:
@jeriox `base` and `poll_base` are rather similar now. To what degree should both be merged?

jeriox (approved):

SilvanVerhoeven:
> > @jeriox `base.html` and `poll_base.html` are rather similar now. To what degree should both be merged?
> 
> The base_poll introduces some logic with respect to voting periods etc. I currently don't see how we could merge that with the base.html without duplicating the logic into the different poll types. If you have an idea feel free to make a suggestion, less code is always better :D

It might be reasonable to introduce a `sidebar_base` and `fullwidth_base` or similar, both based on `base.html`. `sidebar_base` would combine the outer tags and page title from `information_page`, `poll_base` and similar, while `fullwidth_base` serves e.g. `minutes_list`.

Additionally, the page title markup could be moved into its own template

PR #597, opened 2024-07-30
Lint templates

### LINKED ISSUE #580: Integrate HTML formatter
Integrate the [DjHTML](https://github.com/rtts/djhtml) formatter into the project to ensure a common code style when it comes to Django HTML.

_Originally posted by @jeriox in https://github.com/fsr-de/myHPI/pull/525#discussion_r1500844271_

### LINKED ISSUE #586: Add template linter
https://github.com/rtts/djhtml

jeriox:
closes #586 closes #580


jeriox:
@SilvanVerhoeven I removed prettier, as it kept messing with the indention that djhtml did and created an endless loop of reformatting. Could you please have a look if you can restrict prettier to only handle pure js files?

.pre-commit-config.yaml
```diff
@@ -1,8 +1,8 @@
 default_language_version:
-  python: python3.12
+  python: python3.11
 repos:
```
frcroth, line 3: What is the reason for this change?

myhpi/core/templates/core/minutes_list.html
```diff
@@ -1,86 +0,0 @@
-{% extends "base.html" %}
-{% load wagtailcore_tags %}
-{% load i18n %}
```
frcroth, line 3: What has happened here? Diff is showing this file as deleted?

myhpi/static/js/print_processor.js
```diff
@@ -1,67 +1,69 @@
-
 function processPageForPrinting() {
-    moveLinksToFooter();
-    expandAbbreviations();
+  moveLinksToFooter()
+  expandAbbreviations()
```
frcroth, line 3: There seems to have been a previous practice which is overwritten with the new prettier config. Any reason for that change (e.g. now removing semicolons where previously they were ubiquitous?)

myhpi/core/templates/core/minutes_list.html
frcroth: (or at least all lines deleted)

jeriox: yep djhtml crashed when processing this file because of the hardcoded emojis. fixed now

.pre-commit-config.yaml
jeriox: had no 3.12 locally, commit was unintentional. removed now

myhpi/static/js/print_processor.js
jeriox: i have no clue about prettier, as noted above I disabled it anyways because it kept messing with djhtml. Maybe @SilvanVerhoeven or @dropforge can shed some light on that

SilvanVerhoeven:
> I removed prettier, as it kept messing with the indention that djhtml did and created an endless loop of reformatting. Could you please have a look if you can restrict prettier to only handle pure js files?

Seems like we try to lint the same file types here. I think we need to make a decision which formatter should do what. At least for JS I have a slight preference for prettier, as that has proper IDE support, thus allows format on save.

grafik

Maybe [DjLint](https://www.djlint.com/docs/formatter/) could be the perfect fit? It seems to have support for django plus IDE extensions

myhpi/static/js/print_processor.js
SilvanVerhoeven: I think there is no objective reason for the switch - we just had barely anyone writing JS and I prefer without semicolons (cleaner, why write something you don't need). In February, if not even earlier, I used that style and you asked me to share the JS formatter config, so we added prettier yesterday

SilvanVerhoeven:
@jeriox This also closes #580, correct?

jeriox:
> @jeriox This also closes #580, correct?

oh yeah, those are duplicates. I thought we didn't have an issue for it yet and the search also came up empty when I searched for "template" and "linter" ^^

jeriox:
> Seems like we try to lint the same file types here. I think we need to make a decision which formatter should do what. At least for JS I have a slight preference for prettier, as that has proper IDE support, thus allows format on save.
> 

In my understanding from monday we agreed that prettier should only lint pure .js files and djhtml should lint the templates. I strongly favor djhtml for linting the templates as it properly understands django syntax and lints inline js and css. Djhtml does not lint .js files. 

> Maybe [DjLint](https://www.djlint.com/docs/formatter/) could be the perfect fit? It seems to have support for django plus IDE extensions

I have no experience with DjLint and am used to djhtml, but that shouldn't be the problem here. But I'm not sure whether this solves the problem at hand, which is prettier messing with the templates, as DjLint also only lints templates according to their docs. 
On the IDE integration part: I personally hate such "format on save" tools as I don't want them to mess with my code all the time while developing. Formatting before commiting is the better approach IMO, but that is probably just my personal opinion. (and a also don't want to install dozens of extensions in my IDE if they are available in the first place)
Running a specified command (such as djhtml) on save is not possible with your IDE, just extensions?
just to be clear, I don't want to oppose a tool that supports it, it just doesn't convince me to switch to another tool just for that reason

SilvanVerhoeven:
> In my understanding from monday we agreed that prettier should only lint pure .js files and djhtml should lint the templates.

I understood the same :) We never talked about `.css`/`.scss` files, tho. I feel like Prettier is the more common choice here, do you agree?

> Djhtml does not lint .js files.

I just tested it &ndash; due to `id: djjs`, it does. Same for `id: djcss`. Removing both does not affect formatting of template files.

> But I'm not sure whether this solves the problem at hand, which is prettier messing with the templates.

I cannot recreate this &ndash; with the following config, `.html` do not seem to be linted by prettier (as I would expect, since `.html` files are not listed in `tyoes_or`).

```  - repo: https://github.com/rtts/djhtml
    rev: "3.0.6"
    hooks:
      - id: djhtml
  - repo: local
    hooks:
      - id: prettier-eslint
        name: Prettier and ESLint
        entry: prettier-eslint --write --list-different
        language: node
        types_or: [javascript, css, markdown, yaml]
        additional_dependencies: ["prettier-eslint-cli@8.0.1"]
```

> I personally hate such "format on save" tools as I don't want them to mess with my code all the time while developing. 

It makes writing JS much easier for me, as I don't need to care about indentation/can let the IDE clean up everything after I wrote what's necessary.

> Don't want to install dozens of extensions in my IDE if they are available in the first place.

I would never argue against the pre-commit hook, it is very good :) But I'd very much appreciate the option to have it format on save.

>  Running a specified command (such as djhtml) on save is not possible with your IDE, just extensions?

Only after installing an extension for that and configuring it to run a command &ndash; and it seems running that command in a venv is a further headache.

> DjLint only lints templates according to their docs.

Mhh, to me it seems like they also have a [formatter](https://www.djlint.com/docs/formatter/)?  
I think I haven't entirely understood the disadvantages of DjLint, except for missing experience. Could you reiterate this again, please? :)

SilvanVerhoeven:
@jeriox

So my proposal is this:

- use djLint for template files
- use prettier for js, (s)css, md and yaml
- see [configuration](https://github.com/fsr-de/myHPI/pull/597/commits/2b0c4c6f296264fb119daa98ec90924fd915cd07) and [changes on our current code base](https://github.com/fsr-de/myHPI/pull/597/commits/b9b5abf2df843f8c8c45bb2606d94781abd64576)

Advantages:

- djLint formats code, not only indents (djHTML only indents) &rarr; ensures more consistent code style; especially helpful in this project with so many contributors with different backgrounds
- djLint seems to detect more errors &ndash; I found several issues in out code base like missing closing tags after using it
- djLint has VSCode plugin support &ndash; can be easily used on save and shows warnings in the IDE, with the same configuration as used in the pre-commit hook

What do you think of this setup?

myhpi/core/templates/core/minutes.html
```diff
@@ -33,18 +29,14 @@ <h1 class="side-panel-title">{% translate "Minutes taker" %}</h1>
                 <aside class="side-panel border-accent">
                     <h1 class="side-panel-title">{% translate "Participants" %}</h1>
                     <ul>
-                        {% for participant in page.participants.all %}
-                            <li>{{ participant.get_full_name }}</li>
-                        {% endfor %}
+                        {% for participant in page.participants.all %}<li>{{ participant.get_full_name }}</li>{% endfor %}
```
jeriox, line 32: I think having the loop body in the same line makes it much more harder to read it. Can we change that behaviour?

jeriox (commented):
apart from the loops I'm fine with this setup (can't approve as this is my PR)

myhpi/core/templates/core/minutes.html
SilvanVerhoeven: No ideal solution here, I am afraid. We have two options:

1. Reduce `max_line_length` to a smaller value, like `80` (current default `120`). This would break long (i.e. difficult to read) blocks earlier, making your example easier to read. But there are still shorter (i.e. easier to read) `for` blocks that would be in one line.
2. Set `line_break_after_multiline_tag=true`. While this consistently breaks all `for` blocks, it also breaks *every* other tag that *could* contain content over multiple lines. Example: `<i>{% translate "No guests" %}</i>` would be formatted to

```
<i>
    {% translate "No guests" %}
</i>
```

Those unnecessary breaks would happen *a lot*.

That's why I would go with 1. You can see the changes to the code base in the last commit

jeriox: Sad that there is no better option :(

The shorter line length looks okay, although this also produces _some_ instances of the example you gave in 2. This made me think whether we should prefer 2. for a more consistent approach. But I'd also be fine with 1.

@frcroth any thoughts?

SilvanVerhoeven: It could, but only if the line is already very long/complicated. Plus the outer tags usually have more attributes if the line is long.

So, I think it's still reasonable to go for 1, since in practice it keeps the better balance most of the time

frcroth: I am more in favor of 1 but no strong opinion

SilvanVerhoeven (approved):

PR #607, opened 2024-08-08
Ignore lists with missing hashes on the dashboard

lukasrad02:
In our production instance, there are some mailing lists that have not been created with tenca and thus are lacking the tenca hash id. This causes an error on the server when somebody who is member of such a list visits the dashboard.

This PR sets a new flag when querying the lists from tenca to prevent the error to be raised.

### DIFF
--- a/myhpi/tenca_django/views.py
+++ b/myhpi/tenca_django/views.py
@@ -29,7 +29,10 @@ class TencaDashboard(LoginRequiredMixin, FormView):
     def get_context_data(self, **kwargs):
         email = self.request.user.email
         kwargs.setdefault(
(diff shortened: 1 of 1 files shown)


jeriox:
How about adding a hash to those lists? As of #582 we shouldn't have the need to create lists out of tenca anymore?

lukasrad02:
I could imagine there might still be cases in which we want to create lists without involving tenca (internal testing or limitations other than the configurable footer), especially since there are some migrations to the underlying mailing infrastructure planned.

Nevertheless, I will try adding a hash to the existing lists (that just lacked one due to the footer stuff) as soon as the PR you've referenced is in production.

jeriox (approved):

PR #608, opened 2024-08-08
fix: hot fix toc button display with attachments and no toc

### LINKED ISSUE #606: Attachments are missing in mobile view

SilvanVerhoeven:
Closes #606 
Should eventually be fixed properly by #609

### DIFF
--- a/myhpi/core/templates/core/sidebar.html
+++ b/myhpi/core/templates/core/sidebar.html
@@ -40,7 +40,7 @@
     {% endif %}
 </div>
 
-{%if parsed_md.1|hasTocContent %}
+{%if parsed_md.1|hasTocContent or page.attachments.all %}
     <div class="offcanvas offcanvas-bottom" tabindex="-1" id="sidebar-offcanvas" aria-labelledby="sidebar-offcanvas-label">
         <div class="offcanvas-header">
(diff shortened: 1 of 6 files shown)


myhpi/core/templates/core/sidebar.html
```diff
@@ -40,7 +40,7 @@ <h1 class="side-panel-title">{% translate "Attachments" %}</h1>
     {% endif %}
 </div>
 
-{%if parsed_md.1|hasTocContent %}
```
jeriox, line 43: you could also use page.attachements.exists, but its also fine like this

myhpi/core/templates/core/toc_button.html
```diff
@@ -4,12 +4,12 @@
 {% load i18n %}
 {% load bootstrap_icons %}
 
-{%if toc|hasTocContent %}
+{%if toc|hasTocContent or attachments %}
     <div id="sidebar-toggle" class="d-grid d-lg-none d-print-none">
         <button type="button" data-bs-toggle="offcanvas" data-bs-target="#sidebar-offcanvas"
-            aria-controls="sidebar-offcanvas" class="btn btn-light mb-3" aria-label="{% translate 'Table of contents' %}">
```
jeriox, line 10: you could also customize the text to only show what is actually there (just "ToC" if there are not attachements)

jeriox (approved):

PR #621, opened 2024-09-10
integrate tenca code into myHPI

jeriox:
Because of better maintainability and expandability, we wanted to integrate the tenca code into myHPI.

ToDo:

- [ ] create view that serves templates
- [ ] move tests
- [ ] do actual testing with mailman


frcroth:
Would an alternative be to move/fork tenca to the FSR org? Keeping this project small was one of the design goals at the beginning IIRC

lukasrad02:
The rationale behind this decision was that it is cumbersome to apply changes in two repos that depend on each other if there is no need to separate the code. tenca is a very specialized library and most likely nobody but us is using it. So there is no actual benefit from publishing it as a separate library.

Adding a new feature to our mailing lists (for example #582 that I did recently) currently requires you to set up an editable installation of tenca in you venv, apply your changes to the tenca and myHPI source, then have your changes at tenca reviewed, wait for a new tenca version to be released, until you can finally submit a ready-to-merge PR to myHPI with the new tenca version in the dependencies. Including tenca in the myHPI source tree makes this much easier.

myhpi/tenca_django/templates/mailman/unsubscription_message.html
```diff
@@ -0,0 +1,9 @@
+Dear user,
+
+Please confirm your *unsubscription* from {{fqdn_listname}}:
+{{action_link}}
+
+If you did not request this, please report abuse by visiting the following link:
+$action_abuse_link
```
abc013, line 7: Why is this not in `{{ }}`?

myhpi/tenca_django/mailinglist.py
```diff
@@ -0,0 +1,362 @@
...
+			else:
+				raise e
+
+	def toggle_membership(self, email):
+		"""Adds member if not present and removes if already member.
+		
+		This is useful from interfaces where you don't want to hint attackers
+		the current membership status of others.
+
+		Returns a tuple of (status, token), where token is True if the user was newly
```
abc013, line 97: Description incorrect: `status` is set to `True` oder `False`

myhpi/tenca_django/templates/mailman/unsubscription_message.html
jeriox: forgot to replace from the old syntax ^^

myhpi/tenca_django/mailinglist.py
```diff
@@ -0,0 +1,360 @@
...
+
+	def configure_single_template(self, mailman_template_name, tenca_template_name):
+		template_args = dict(
+			fqdn_listname=self.fqdn_listname,
+			action_link=self.build_action_link('$token'),
+			action_abuse_link=self.build_action_abuse_link('$token'),
+			invite_link=self.build_invite_link(),
+			web_ui=urllib.parse.urljoin(settings.GET_SITE_URL(), reverse("tenca_dashboard"))
+		)
+		self.list.set_template(mailman_template_name, urllib.parse.urljoin(settings.SITE_URL, reverse("tenca_django:mailman_template", kwargs={'template_name': tenca_template_name})))
```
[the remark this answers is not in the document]
lukasrad02: Currently, the `template_args` are query parameters. However, mailman does not care about the URL format, so using a path-structure (e. g. `…/<fqdn-listname>/<action-link>/…`) or anything else would be okay, too. Old template links will break anyways and we have to change them all (as they point to `localhost:8080` and we're moving the templateserver to `myhpi.de`), so changing the URL format would not cause any additional effort.

Instead of inserting `fqdn_listname` into the domain on our own, we could also use the mailman placeholder `$listname` (<https://docs.mailman3.org/projects/mailman/en/latest/src/mailman/rest/docs/templates.html#url-placeholders>).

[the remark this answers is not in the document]
lukasrad02: Fixed in f2f4d938e00f4600f8fbf7e201f712b88a3d9f90

PR #629, opened 2024-10-22
added machine translated hint to information_page.html

geromequa:
added machine translated hint to information_page.html

### DIFF
--- a/myhpi/core/templates/core/information_page.html
+++ b/myhpi/core/templates/core/information_page.html
@@ -16,6 +16,12 @@
     </div>
     <div class="col-lg-3">
         <div class="side-panel-container">
+            {% get_language_info for LANGUAGE_CODE as lang %}
+            {% if lang.code == "en" %}
+                <aside class="side-panel border-accent">
+                    This has page has been machine-translated.
+                </aside>
+            {% endif %}
             <aside class="side-panel border-accent">
                 <h1 class="side-panel-title">{% translate "Last edited" %}</h1>
                 {% if not page.last_published_at %}


jeriox:
> added machine translated hint to information_page.html

wouldn't it be more important for minutes? IMO translations of information pages are more carefully checked than minutes translations

lukasrad02:
I would wish this being configurable on a per-page basis, as there are some pages that are actively curated in English

myhpi/core/templates/core/information_page.html
```diff
@@ -16,6 +16,12 @@ <h1 class="page-title">
     </div>
     <div class="col-lg-3">
         <div class="side-panel-container">
+            {% get_language_info for LANGUAGE_CODE as lang %}
+            {% if lang.code == "en" %}
+                <aside class="side-panel border-accent">
+                    This has page has been machine-translated.
```
frcroth, line 22:
```suggestion
                    This page may have been machine-translated.
```

frcroth: I think this would be better since other Gremien may actually manually translate it.

Also we could think about explaining it better. Why is it machine translated? What can I do about it?

frcroth (commented):
Also add for minutes.

myhpi/core/templates/core/information_page.html
frcroth: I agree that for a user, an explicit "is" would be more valuable.With the current specification of the issue I believe "may" is the best solution. This means I will be alerted to possible weirdness and know I will ask if something is unclear.

But if we expand the scope we can also have the is / is not . But I think this definitely needs to be set at the time of translation, automatically. So it may be more complicated, and I think this version would already be helpful (though the note may need to be expanded to explain a little better why the hint is therek

lukasrad02: > But I think this definitely needs to be set at the time of translation

Yes, this should be set when translating a page

> automatically

Only if it it still possible to change this. Otherwise, it would not be possible to machine-translate a page, review it (without any need for changes) and then publish it without a warning.

frcroth: > Only if it it still possible to change this.  Otherwise, it would not be possible to machine-translate a page, review it (without any need for changes) and then publish it without a warning.

Agreed. The question remains whether this PR (in some form, e.g. adapted wording ) is useful or not.

lukasrad02: In my opinion, the following approaches would be useful:

- Use more open wording, e. g. "English pages are typically machine-translated and reviewed, but errors may occur. If you're unsure about some details, feel free to reach out to the page authors". As this is pretty long, maybe just show some "ℹ️ About this translation" entry in the sidebar and use the long text as a tooltip of the info icon
- Use the current wording but make it configurable on a per-page basis whether the notice is shown or not

Issue #612, opened 2024-08-12
Search result highlighting breaks markdown rendering


lukasrad02:
#557 added highlighting of the search term in search results. This could break the rendering of search results if the term occurs in identifiers (e.g. internal links) instead of normal text.

### Steps to reproduce

- Create two information pages. Let x, y be their IDs
- On the first page, create an internal link to the second page like `[Second page](page:y)` and publish this change
- Search for the numeric value of ID y
- There is an internal server error, since `y` has been replaced by `**y**`, which cannot be casted to an int

corvin-koegler:
Hello there!

abc013:
General Kenobi!

jeriox:
btw the ordering of results is also not correct, in case you want to tackle this as well

corvin-koegler:
> btw the ordering of results is also not correct, in case you want to tackle this as well

![image](https://github.com/user-attachments/assets/00c275e4-b6be-41c7-bf73-81f34eabc217)

Looks like they are sorted by creation date. What is the intended behaviour?

jeriox:
![image](https://github.com/user-attachments/assets/77ed10ec-945f-44e4-a4ec-ee4f1acf431e)
not really, they are all over the place. Intended behaviour would be to use the minutes date for minutes. It would probably be best to sort information pages to the front to make them visible

PR #630, opened 2024-10-22
Fixed search page-id edgecase

corvin-koegler:
Adding a space to numeric searches prevents markdown from breaking the search.

[Issue 612](https://github.com/fsr-de/myHPI/issues/612#issue-2460222712)

### DIFF
--- a/myhpi/search/templatetags/search.py
+++ b/myhpi/search/templatetags/search.py
@@ -12,6 +12,9 @@ register = template.Library()
 @register.filter(name="highlight_query_markdown")
 # select 3 lines around first match and highlight query match
 def highlight_query(content, search_query, surrounding_lines=1):
+    # if query coincides with page id the search breaks without the extra space
+    if search_query.isnumeric():
+        search_query = str(search_query) + " "
     lines = content.split("\n")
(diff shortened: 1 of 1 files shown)


frcroth:
Can you explain why this happens?

lukasrad02:
This only fixes the symptom, but not the actual root cause.

The underlying problem of #612 is that the highlighting logic replaces substrings in the markdown text. So, for example, searching for "page" also breaks our internal links (in the form of `[Text](page:123)`). This query does not cause a internal server error, as the markdown renderer just uses `**page**:123` as link target, but it still alters the meaning.

Hence, we either have to use a more sophisticated highlighting logic (parsing the markdown into an AST, then performing replacements only on textnodes and then writing the AST back to markdown), or have to remove the highlighting completely.

lukasrad02:
I've checked out your branch and couldn't see any highlighting at all except for search results in the page title. Did you also experience this issue?

Also, I'm a bit afraid that rendering markdown twice might break under certain circumstances, e.g. escaped characters in the original markdown get converted into the plain character in the first rendering process, but as the markdown is rendered again, it might then be interpreted as some markdown element…

PR #636, opened 2024-11-20
Localize abbreviation explanations

### LINKED ISSUE #634: Abbreviation Localisation
Abbreviations do not respect languages: There are abbreviations, but they do not consider the page language. There should be different abbreviations for different locales.

frcroth:
Fixes #634 

This adds a language field to Abbreviation Explanations.
- Set the language when creating a new abbreviation explanation (en or de)
- Abbreviation explanations will only be shown if the current language matches the one stored in the abbreviation
- Create abbreviations for English and German separately!

image


jeriox:
I would prefer using wagtail-localize for this as well (see https://wagtail-localize.org/stable/how-to/modeladmin/)

jeriox:
> Though this is for ModelAdmin and we are using snippets, so I do not know if this can be used?

good point. Head here for snippets: https://docs.wagtail.org/en/stable/advanced_topics/i18n.html#making-snippets-with-existing-data-translatable

frcroth:
I tested it using TranslatableMixin. However, I am not sure I like the result.
Basically the workflow is that you have AbbreviationExplanations and then create translations for them.
- Have FSR abbreviation
- Create a translation
- Set explanation to something different
- In an english text "FSR" may now be explained as "Student Representative Group"

So this would make it possible to use the same abbreviation in english text with a different explanation. This is certainly a use case. But I don't think this workflow really accounts for english abbreviations. You can also translate the abbreviation itself, which is pretty weird.

jeriox:
IMO a separate english abbreviation can just be added like any ohter german one, as it will only be rendered when it is used in the text. So I wouldn't add code for that. Having an english explanation for german abbreviations (like StuKo) on the other hand is really helpful and what would be achieved with TranslatableMixin

But thats just my opinion, maybe @lukasrad02 wants to chime in with the customer opinion ^^

lukasrad02:
I think our most frequent use case are "fixed" abbreviations that are the same in German and English. For example, we don't use "SRG" in English, but either spell it out as "Student Representative Group" or stick to "FSR". The same applies for all other committees as well es terms like "BAMA-O".

Translated abbreviations also occur sometimes, but are pretty rare. The only case I found when checking the Erstiheft (where we already maintain German and English abbreviations) are credit points, which are abbreviated with "CP" in English, while we'Re using "LP" ("Leistungspunkte") in German.

So I prefer the approach using wagtail-localize.

frcroth:
Now implemented via wagtail-localize. As per the discussion, the abbreviation can not be edited, only the explanation. @jeriox @lukasrad02

lukasrad02 (approved):

jeriox (approved):

PR #642, opened 2025-03-19
Fix coverage report being skipped in CI

frcroth:
This bug was introduced in #489.


lukasrad02:
To prevent this from happening again, I would suggest to add a comment at the python version in the matrix definition that when updating the python version, the condition of the coverage job also has to be updated.

lukasrad02 (approved):

PR #646, opened 2025-03-19
Hide navbar entries that are still in animation

### LINKED ISSUE #633: Drop down menu can have items from multiple menu items
On myHPI, when selecting Fachschaftsrat in the menu and then quickly selecting another entry, e.g. Informationen, both entries' subpages will be in the dropdown menu with both menu items selected (orange underlined).

This is confusing.

frcroth:
Fix #633

Also, I installed the pre commit hook and as I don't have Python 3.12, I deleted that constraint. It should still work for people with Python 3.12.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         python-version: [3.11]
@@ -46,7 +46,7 @@ jobs:
           poetry run djlint --lint myhpi
 
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         python-version: [3.11, 3.12] # Also update Coveralls step when updating versions here
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To set up a development version on your local machine, you need to execute the f
 1. Optionally: Create test data with `python manage.py create_test_data`
 1. Create a local superuser with `python manage.py createsuperuser`
 1. Start the development server with `python manage.py runserver`
-1. Open your web browser, visit `http://localhost:8000/admin` and log in with the user you just created
+1. Open your web browser, visit `http://localhost:8000/admin/` and log in with the user you just created
 
 ### Tests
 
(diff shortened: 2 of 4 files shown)


myhpi/static/js/navbar.js
lukasrad02, line 1: This workaround is flawed, since `isCollapsing` is true for both opening and closing containers. Thus, when you have container A open and then quickly click A and B, everything seems to be fine (A closes, B opens and stays visible). However, the event handler for `shown.bs.collapse` is assigned to A, causing it to auto-close immediately when you open it the next time.

This happens both when A is closed explicitly (by clicking A) as well as when A is closed implicitly (by clicking a third item C).

lukasrad02 (changes requested):

lukasrad02 (approved):

myhpi/static/js/navbar.js
```diff
@@ -79,7 +92,23 @@ const collapseOthersOnSameLevel = (navItemContainer) => {
   navContainersOnSameLevel.forEach((navContainerOnSameLevel) => {
     if (navContainerOnSameLevel === navItemContainer) return
     if (isCollapsed(navContainerOnSameLevel)) return
-    bootstrap.Collapse.getOrCreateInstance(navContainerOnSameLevel).hide()
+    // When clicking on multiple navItems in quick succession, some might not yet be collapsed, but "collapsing".
+    // In that case, we wait for the collapse to finish before hiding them.
```
SilvanVerhoeven, line 96:
```suggestion
    // When opening two navItems in quick succession, the first might not yet be expanded, but in the process of expanding. Bootstrap calls this state "collapsing".
    // Hiding a collapsing item has no effect; the first navItem would stay expanded after finishing. Both navItems would be expanded.
    // So we wait for the collapsing to finish before hiding the item properly.
```

myhpi/static/js/navbar.js
```diff
@@ -79,7 +92,23 @@ const collapseOthersOnSameLevel = (navItemContainer) => {
...
+    if (isCollapsing(navContainerOnSameLevel)) {
+      navContainerOnSameLevel.addEventListener(
+        "transitionend",
+        (event) => {
+          if (event.target.classList.contains("show")) {
```
SilvanVerhoeven, line 101:
```suggestion
          if (isCollapsed(event.target)) {
```

PR #666, opened 2025-07-23
Fix poll preview and localize options

### LINKED ISSUE #527: Improve Poll Page Preview
Currently, the preview of a poll cannot be used to preview a poll including its options:

- When viewing the preview before or after the voting period, one can only see the poll description and the notice *You've accessed this page outside of the voting period.*, but not the poll options. I would wish to be able to see the options in the preview for easier reviews.
- When viewing the preview during the voting period, rendering fails with the error `'FakeQuerySet' object has no attribute 'intersection'` at `{% if page|can_vote:request.user %}`. I would wish for the preview to ignore the permissions of the user and always show the options.

During these changes, we should forbid submitting actual votes if this is not already implemented (couldn't check due to the error mentioned above).

### LINKED ISSUE #542: Fix localization of polls
The polls feature needs better integration into page localizations:

- [ ] Allow localizing options and option descriptions
- [ ] Don't treat translated polls as separate polls
  - [ ] A user who voted on a German poll should not be able to vote on the corresponding English poll and vice versa
  - [ ] Results for a poll should take votes from the German and English variant into account
- [ ] Synchronize properties like the eligible groups (see #373 and the fix in #491)

frcroth:
This PR:
- fixes poll previews
- Allows localization of options
	- These are now actually shown
- Changes poll behavior so that all poll localizations use a canonical poll in the default locale for can_vote and cast_vote
- Preview should see options and not be able to vote.
- Poll properties are kept in sync

For testing:
Create a poll before applying migrations to test the migration.

Fixes #542 
Fixes #527


myhpi/polls/migrations/0014_enforce_translation_key_locale_unique.py
lukasrad02, line 1: It seems like this final migration does not create the exact state that corresponds to the models. At least, my local setup is warning me about missing migrations in the polls app

lukasrad02 (changes requested):
Only some small issues left

lukasrad02:
Oh, I just noticed a regression in your PR: Ordering choices/options does not on both poll types work after applying your migrations.

I suspect this is related to clearing the model options in migration 0012, at least the docs specify that these options are used for ordering (among others) and migration 0011 sets `"ordering": ["sort_order"]`

frcroth:
> Oh, I just noticed a regression in your PR: Ordering choices/options does not on both poll types work after applying your migrations.
> 
> I suspect this is related to clearing the model options in migration 0012, at least the docs specify that these options are used for ordering (among others) and migration 0011 sets `"ordering": ["sort_order"]`

Good catch!
If I understand it correctly, the Meta class of the model was now taken from TranslatableMixin and no longer from Orderable. So I added it explicitly with the two Metas merged.

lukasrad02 (approved):
Lgtm, thank you for addressing all the poll issues!

Issue #614, opened 2024-08-16
Exclude heading anchors in translations


lukasrad02:
#525 introduced permalinks to headings within a page, displayed by a `#` after the heading text. These links are added to the page in a way that they are also shown in the translation editor.

I.e., a heading `# Beispiel` that was previously shown as `Beispiel` in the translation editor will now be shown as `Beispiel<a id="a1">#</a>`.

This causes two problems:

1. Every first sync of a page after the update causes all headings to not be translated, as the heading in the original language has changed from the perspective of the translation editor. This makes updates tedious, but at least only once per page.
2. DeepL does not handle these kinds of links correctly, so machine translated headings look weird (see below).

Examples for bad link handling in automatic translations:
![Example where the full heading has become the permalink](https://github.com/user-attachments/assets/f2af31c2-5136-4d7a-8f18-6685c7638457)
![Example where one single word within the heading has become the permalink](https://github.com/user-attachments/assets/b4f166c6-8cca-4c69-98b4-487888bcc4dc)

In practice, this does not seem to be an issue, as the broken translations shown above are rendered correctly (the actual heading is not a link, there is no additional `#`, but the permalink-`#` is visible on hover) and one can even remove the anchor tag from the translated heading completely without removing the permalink from the rendered page.

Nevertheless, to avoid confusion and resolve the issue regarding the first sync of existing pages, it would be nice if we could remove the `<a id="a1">#</a>` part from the original headings in the translation editor.

PR #674, opened 2026-02-12
Update dependencies to fix machine translation

just-spafi:


lukasrad02:
> we could just update wagtail localize.

Yes, that's what I wanted to imply

poetry.lock
lukasrad02, line 1: The lockfile seems to still reflect the full upgrade of all dependencies and not only the `wagtail-localize` update

lukasrad02 (dismissed):

poetry.lock
just-spafi: Should be fixed now

lukasrad02:
Even after 0ebfbc46cd38181d3772e733c6d1aa9a22e78a3c, the lockfile still contains more changes than it should, but I won't care about this, since no major versions have been changed, so everything should still be compatible.

However, I noticed a bug regarding translations after the update: Clicking the "Translate" button activates the text box for the wrong snippet. Working with the dummy data (`python manage.py create_test_data`), when I try to translate the page title of the `/de/information/first-semester/` page, one of the links at the bottom gets edited instead:

[myhpi.webm](https://github.com/user-attachments/assets/df13f0ef-1f56-4729-9171-e081a758b9d3)

I think this is related to #614, since the translated string is inserted into the heading anchor on the page

styrix560 (approved):
did not test, but code looks good given that we only reverted one commit and updated some other stuff.

Belissimo-T (approved):
looks good to me
I don't get why half of the displayed changes are poorly formatted layout/whitespace diffs

PR #675, opened 2026-02-17
Switch from PostgreSQL-action to  PostgreSQL service configuration in test workflow

just-spafi:
As noticed in #674, the PostgreSQL-action was not updated for 7 years and is now incompatible with github actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,6 +47,19 @@ jobs:
 
   test:
     runs-on: ubuntu-24.04
+    services:
+      postgres:
+        image: postgres:11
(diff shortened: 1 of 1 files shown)


Belissimo-T (approved):
A direct replacement could also be https://github.com/ikalnytskyi/action-setup-postgres.

I feel really underqualified to review this, but it looks good to me.

just-spafi:
> A direct replacement could also be https://github.com/ikalnytskyi/action-setup-postgres.

Could be, but right now we use Postgres 11 and this action only supports Postgres 14 and newer.

